Issue #28, opened 2023-08-28
Integrate Anthropic Claude as an additional LLM


ibuchh:
I believe we have an opportunity to elevate this project even further by considering integration with Anthropic Claude Language Model (LLM). With my prior experience, I can create the necessary code to seamlessly incorporate Anthropic Claude into our project. This integration would undoubtedly bolster our project's capabilities and offer our users a comprehensive and powerful experience.

ibuchh:
Yes indeed.

ishaan-jaff:
@ibuchh @massi-ang you can do this very easily using LiteLLM https://github.com/BerriAI/litellm/

it allows you to use any llm as a drop in replacement for gpt-3.5-turbo

dairiley:
Anthropic Claude is available via Bedrock.

PR #31, opened 2023-09-05
Add support for Azure OpenAI, Palm, Replicate, Sagemaker (100+LLMs) - using litellm

ishaan-jaff:
*Issue #, if available:*
https://github.com/aws-samples/aws-genai-llm-chatbot/issues/28

*Description of changes:*
This PR adds support for 50+ models with a standard I/O interface using: https://github.com/BerriAI/litellm/

`ChatLiteLLM()` is integrated into langchain and allows you to call all models using the `ChatOpenAI` I/O interface 
https://python.langchain.com/docs/integrations/chat/litellm

Here's an example of how to use ChatLiteLLM()
```python
ChatLiteLLM(model="gpt-3.5-turbo")
ChatLiteLLM(model="claude-2", temperature=0.3)
ChatLiteLLM(model="command-nightly")
ChatLiteLLM(model="replicate/llama-2-70b-chat:2c1608e18606fad2812020dc541930f2d0495ce32eee50074220b87300bc16e1")

```

### DIFF
--- a/lib/model-interfaces/langchain/functions/request-handler/adapters/openai/gpt.py
+++ b/lib/model-interfaces/langchain/functions/request-handler/adapters/openai/gpt.py
@@ -1,6 +1,6 @@
 import os
 
-from langchain.chat_models import ChatOpenAI
+from langchain.chat_models import ChatLiteLLM
 
 from ..base import ModelAdapter
 from ..registry import registry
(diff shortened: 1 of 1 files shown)


ishaan-jaff:
@bigadsoleiman @spugachev can I get a review on this PR ? 

happy to add more docs/tests if this initial commit looks good

bigadsoleiman:
Hi @ishaan-jaff thanks for the submission.

I think it's good addition to provide an adapter for `ChatLiteLLM` but rather than replacing the GPT one I would rather have a new one specifically for ChatLiteLLM explicitly named `ChatLiteLLMAdapter` to avoid confusion

massi-ang:
Hi Ishaan, 

your project looks really interesting, but as you might have seen, this solution relies on Langchain not only for the LLM model adapters but also for additional functionality like Memory, Chains, Prompts, etc. I find that your library could make it easier for tools like Langchain to onboard multiple LLM in an easier way, but at them moment we are not in the position to introduce it in our solution as is since it would require to also code the additional langchain functionality.

PR #51, opened 2023-09-27
Added _language_code attribute filter to AmazonKendraRetriever

cmilanf:
*Description of changes:*
`rag-source` for KENDRA only supports English language indexed documents due to the lack of interface to configure an `attribute_filter` in `AmazonKendraRetriever()`. While the ideal solution would be to add language configuration in the UI, this is a stopgap solution that adds `KENDRA_LANGUAGE_CODE` environment variable to the `api-handler` Lambda function, that will be subsequently used as an _attribute_filter_ in `AmazonKendraRetriever` call. Without any `attribute_filter` in the call, KENDRA defaults to discard non-English documents.

To avoid any any disruption in the current solution behaviour, `KENDRA_LANGUAGE_CODE` defaults to `en`, so users of the sample won't notice any change unless explicitly change variable value.

### DIFF
--- a/lib/rag-sources/aurora-pgvector/functions/document-indexing/requirements.txt
+++ b/lib/rag-sources/aurora-pgvector/functions/document-indexing/requirements.txt
@@ -1,6 +1,6 @@
 aws_lambda_powertools
 aws_xray_sdk
-langchain
+langchain==0.0.288
 psycopg2-binary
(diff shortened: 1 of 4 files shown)


cmilanf:
Closing this PR since it was intended for v2, not being applicable to the recently released v3.

Issue #52, opened 2023-09-27
Fix Bedrock Anthropic Integration


FabioChiodini:
Given that the invocation of Anthropic has changed in Bedrock could you please fix it to make it work again?

"We are requiring all Amazon Bedrock customers consuming Anthropic’s Claude to send prompts to the model in the following syntax: “\n\nHuman: <Customer Prompt>\n\nAssistant:”."

Now the invocation of Claude fails with this error:

"<class 'ValueError'>:Error raised by bedrock service: An error occurred (ValidationException) when calling the InvokeModel operation: Invalid prompt: prompt must end with "

Assistant:" turn"

Thank you very much.

massi-ang:
This is addresses by #55

PR #53, opened 2023-09-27
fix: Anthropic prompt (#52)

massi-ang:
*Issue #, if available:*
#52 

*Description of changes:*
- adding a new get_prompt function to get the question prompt
- override default prompts on Rag

### DIFF
--- a/lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
+++ b/lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
@@ -2,7 +2,7 @@ import os
 
 import boto3
 from langchain.agents import ZeroShotAgent
-from langchain.llms import Bedrock
+from langchain.chat_models import Bedrock
 from langchain.prompts.prompt import PromptTemplate
 
 from ..base import ModelAdapter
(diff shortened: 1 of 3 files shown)


FabioChiodini (approved):
LGTM

lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
```diff
@@ -2,7 +2,7 @@
 
 import boto3
 from langchain.agents import ZeroShotAgent
-from langchain.llms import Bedrock
+from langchain.chat_models import Bedrock
```
bigadsoleiman, line 5: Not sure about this line...

Looks like [chat_models](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/chat_models/__init__.py#L24) exposes only `BedrockChat` and not `Bedrock` ?

massi-ang: Good catch! It should import from `llms` and not `chat_models`

bigadsoleiman (approved):

Issue #55, opened 2023-09-28
Allow overriding of the `ConversationalRetrievalChain` prompts


massi-ang:
`ConversationalRetrievalChain` is used for RAG and we want to allow overriding of the Langchain default prompts. This solves issues with prompt specific patterns (eg Claude) but also for fine tuning of the prompts to achieve better results.

PR #57, opened 2023-09-28
feat: possibility to override the Retrieve prompts

massi-ang:
*Issue #, if available:*
#55 

*Description of changes:*
RAG uses langchain `ConversationalRetrievalChain` and with this change we allow to override the 2 prompts used by this chain this fixes the issue with the specific Claude formatting but also allows for further prompt tuning if necessary

### DIFF
--- a/lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
+++ b/lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
@@ -1,8 +1,7 @@
 import genai_core.clients
 
+from langchain.llms import Bedrock
 from langchain.prompts.prompt import PromptTemplate
-
-from langchain.chat_models import BedrockChat
 
 from ..base import ModelAdapter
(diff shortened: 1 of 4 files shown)


lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/claude.py
```diff
@@ -25,23 +24,40 @@ def get_llm(self, model_kwargs={}):
         if "maxTokens" in model_kwargs:
             params["max_tokens_to_sample"] = model_kwargs["maxTokens"]
 
-        return BedrockChat(
-            client=bedrock,
+        return Bedrock(
+            client=client,
```
bigadsoleiman: looks like client is not defined above (it's still `bedrock` if you were looking to rename it)

bigadsoleiman: I missed the next commit, all good.

PR #96, opened 2023-10-09
Allow to access local Kendra Indexs without role

spugachev:
Allow to access local Kendra Indexs without roleArn


spugachev:
Allow to access local Kendra Indexs without roleArn

Issue #215, opened 2023-11-13
Feature request: add support for LLama2 13B Chat on Bedrock


bigadsoleiman:
- Add adapter for [Llama Chat 13B](https://aws.amazon.com/about-aws/whats-new/2023/11/llama-2-chat-13b-model-meta-amazon-bedrock/) on Bedrock

PR #221, opened 2023-11-16
Add Llama2 support for Amazon Bedrock. Fix prompts for Llama2 on SageMaker. 

bigadsoleiman:
Related to #215 

- Local override of `BedrockBase` until https://github.com/langchain-ai/langchain/issues/13316 is released only for LLama2 adapter for Bedrock.
- Add `shared` folder with common assets between providers
- Add shared PromptTemplates and Custom `ConversationBufferMemory` for LLama2
- Update existing SageMaker `SMLlama2ChatAdapter` to use shared prompt templates and ConversationBufferMemory
- Add LLama2 adapter for Bedrock consuming shared prompt templates and ConversationBufferMemory
- Remove LLama2 Base adapter as no longer in use
- Improve adapter error message when adapter is not available.
- Ran black for prettifying pyhon code.


lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
```diff
@@ -0,0 +1,64 @@
...
+from langchain.prompts import PromptTemplate
+
+
+Llama2ChatPrompt = """[INST] <<SYS>>
+You are an helpful assistant that provides concise answers to user questions with as little sentences as possible and at maximum 3 sentences. You do not repeat yourself. You avoid bulleted list or emojis.
+<</SYS>>
+
+{chat_history}
+
+{input} [/INST]"""
```
massi-ang, line 14: When using this prompt with the SM endpoint, the resulting prompt sent to the LLM becomes 
```
<s> [INST] [INST] <<SYS>>
You are an helpful assistant that provides concise answers to user questions with as little sentences as possible and at maximum 3 sentences. You do not repeat yourself. You avoid bulleted list or emojis.
<</SYS>>
{chat_history}
{input} [/INST] [/INST]
```

I would recommend that you remove the initial and final `[INST]` tags from the prompt and add them instead in the `transform_input` function. Alternatively, leave the promps as is and modify the `transform_input` function to remove them.

lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_buffer_string(self) -> str:
+        """modified version of https://github.com/langchain-ai/langchain/blob/bed06a4f4ab802bedb3533021da920c05a736810/libs/langchain/langchain/schema/messages.py#L14"""
+        human_message_cnt = 0
+        string_messages = []
+        for m in self.chat_memory.messages:
+            if isinstance(m, HumanMessage):
+                if human_message_cnt == 0:
+                    message = f"{m.content} [/INST]"
+                else:
+                    message = f"[INST] {m.content} [/INST]"
```
massi-ang, line 56: this should be `f"<s> [INST] {m.content} [/INST]"`

lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
```diff
@@ -0,0 +1,64 @@
...
+                human_message_cnt += 1
+            elif isinstance(m, AIMessage):
+                message = f"{m.content}"
```
massi-ang, line 59: this should be `f"{m.content} </s>"`

lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
massi-ang: As you can see here https://github.com/facebookresearch/llama-recipes/blob/38df368a70c292db6e42e0f275139678497fd896/src/llama_recipes/inference/chat_utils.py#L45-L47 the code used by Langchain tokenizes each pair or human/assistant dialogs individually and then concatenates them at the end. Each tokenization add the `<s>` token at the beginning of the array `[1, ...]`, while the code explicitly adds the `</s>` token at the end `+ [tokenizer.eos_token_id]` resulting in `[1,...,2]`. Since we are passing a string an not an array of dialogues, the content is tokenized only once, and we need to add all the intermediate `<s>` and `</s>` tokens that would otherwise be generated by their code.

lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
massi-ang: see explanation above

lib/model-interfaces/langchain/functions/request-handler/adapters/sagemaker/mistralai/mistral_instruct.py
```diff
@@ -58,7 +58,7 @@ def get_llm(self, model_kwargs={}):
             model_kwargs=params,
             callbacks=[self.callback_handler],
         )
-    
+
     def get_qa_prompt(self):
         template = """<s>[INST] Use the following pieces of context to answer the question at the end. If you don't know the answer, just say that you don't know, don't try to make up an answer.[/INST]
```
massi-ang, line 63: Not related to this PR, but it could `<s>` should be removed. Mistral prompts should be identical to LLama2 Chat SM prompts

lib/model-interfaces/langchain/functions/request-handler/adapters/shared/meta/llama2_chat.py
```diff
@@ -0,0 +1,64 @@
...
+
+{input} [/INST]"""
+
+Llama2ChatQAPrompt = """<<SYS>>
+Use the following conversation history and pieces of context to answer the question at the end. If you don't know the answer, just say that you don't know, don't try to make up an answer. You do not repeat yourself. You avoid bulleted list or emojis.
+<</SYS>>
+
+{chat_history}
+
+[INST] Context: {context} [/INST]
```
massi-ang, line 22: This look strange and would not follow the pattern. I would just remove `[INST]` and `[/INST]` from this line.

massi-ang (approved):
Two minor comments.
Good to go.

PR #240, opened 2023-11-23
fix(create_tool): better hints on multiselect

massi-ang:
*Issue #, if available:*

*Description of changes:*
This is an effort to improve the user experience by providing clearer and emphasized instructions on how to use the multiselect option in the configuration creation tool.

image

NOTE: blue highlight is just to show the difference

image

### DIFF
--- a/cli/magic-create.ts
+++ b/cli/magic-create.ts
@@ -156,8 +156,8 @@ async function processCreateOptions(options: any): Promise<void> {
     {
       type: "multiselect",
       name: "sagemakerModels",
-      message:
-        "Which SageMaker Models do you want to enable (enter for None, space to select)",
+      hint: "SPACE to select, ENTER to confirm selection",
+      message: "Which SageMaker Models do you want to enable",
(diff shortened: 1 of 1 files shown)


cli/magic-create.ts
```diff
@@ -156,8 +156,8 @@ async function processCreateOptions(options: any): Promise<void> {
     {
       type: "multiselect",
       name: "sagemakerModels",
-      message:
-        "Which SageMaker Models do you want to enable (enter for None, space to select)",
+      hint: "SPACE to select, ENTER to confirm selection",
```
bigadsoleiman, line 159: Can we add "Optional" so it doesn't look like you have to choose one?

massi-ang: What about `"Do you want to enable any SagemMaker models?"`

bigadsoleiman: Yes

bigadsoleiman (approved):

PR #250, opened 2023-11-28
fix(genai_core): Added handling for bad FMs

flamingquaks:
Bedrock can return FMs that are not valid, such as `foundation-model/amazon.dummy-model`

*Description of changes:*
Currently, when calling list models in boto3 for bedrock, there is an exception caused if the list of models returns invalid models. 
An example of a bad model in the list of models:
`{'modelArn': 'arn:aws:bedrock:us-west-2::foundation-model/amazon.dummy-model', 'modelId': 'amazon.dummy-model', 'inferenceTypesSupported': ['ON_DEMAND']
    }`

Since this model has no `inputModalities` or `outputModalities` it raises an exception and returns no models back. 

This change adds checks for `inputModalities` and `outputModalities` being present in the model details and skips the model if it's invalid. 

This change has been applied to both the `list_bedrock_finetuned_models` & `list_bedrock_models`

A change was also `list_bedrock_finetuned_models` to change the raised exception to differentiate exceptions between `list_bedrock_models` and `list_bedrock_finetuned_models`

### DIFF
--- a/lib/shared/layers/python-sdk/python/genai_core/models.py
+++ b/lib/shared/layers/python-sdk/python/genai_core/models.py
@@ -70,8 +70,10 @@ def list_bedrock_models():
             }
             for model in bedrock_models
             # Exclude embeddings and stable diffusion models
-            if Modality.EMBEDDING.value not in model["outputModalities"]
-            and Modality.IMAGE.value not in model["outputModalities"]
+            if model.get("inputModalities", None) != None
+            and model.get("outputModalities", None) != None
+            and Modality.EMBEDDING.value not in model.get("outputModalities", [])
+            and Modality.IMAGE.value not in model.get("outputModalities", [])
         ]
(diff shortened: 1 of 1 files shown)


lib/shared/layers/python-sdk/python/genai_core/models.py
```diff
@@ -70,8 +70,10 @@ def list_bedrock_models():
             }
             for model in bedrock_models
             # Exclude embeddings and stable diffusion models
-            if Modality.EMBEDDING.value not in model["outputModalities"]
-            and Modality.IMAGE.value not in model["outputModalities"]
+            if model.get("inputModalities", None) != None
```
massi-ang, line 73: prefer a simpler `"inputModalities" in model`

bigadsoleiman (approved):

PR #251, opened 2023-11-30
Fix error with module not found for feedparser

agscripter:
*Issue #, if available:*
There is no issue available.

*Description of changes:*
* Adding `feedparser==6.0.10` to the list of requirements for a task.
* It fixes the following error: `ModuleNotFoundError: No module named 'feedparser'`
* It happens when processing files for workspaces in the ECS docker image


flamingquaks:
I reviewed this and confirm this bug was introduced as part of the RSS subscription integration. 

I've searched through the code base for other possible related issues, but I think this was the only one.

bigadsoleiman (approved):
Thanks guys, merging.

Issue #267, opened 2023-12-13
Display only active Bedrock models


bigadsoleiman:
Leverage `FoundationModelLifecycle` to filter out `LEGACY` models. [API Ref](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_FoundationModelLifecycle.html)

massi-ang:
@bigadsoleiman Hi Bigad, are you working on this?

bigadsoleiman:
Not yet

PR #275, opened 2023-12-15
Feature: filter out Bedrock models that cannot be used in the playground

massi-ang:
*Issue #, if available:*

resolves: #267 

*Description of changes:*
Only retrieve ON_DEMAND inference models that support TEXT output when listing foundation models, and filters out those that are not marked as ACTIVE. For this to work `boto3` was updated to latest version.

### DIFF
--- a/lib/shared/layers/python-sdk/python/genai_core/models.py
+++ b/lib/shared/layers/python-sdk/python/genai_core/models.py
@@ -55,8 +55,14 @@ def list_bedrock_models():
         if not bedrock:
             return None
 
-        response = bedrock.list_foundation_models()
-        bedrock_models = response.get("modelSummaries", [])
+        response = bedrock.list_foundation_models(
+            byInferenceType="ON_DEMAND", byOutputModality="TEXT"
+        )
+        bedrock_models = [
+            m
+            for m in response.get("modelSummaries", [])
+            if m.get("modelLifecycle", {}).get("status") == "ACTIVE"
+        ]
 
         models = [
             {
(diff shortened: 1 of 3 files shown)


QuinnGT:
Great timing! I was working on this last night and only tested the status but didn't think to test the inference type. Just tested your fix and it works well. Thanks!

lib/shared/layers/python-sdk/python/genai_core/models.py
```diff
@@ -55,8 +55,14 @@ def list_bedrock_models():
         if not bedrock:
             return None
 
-        response = bedrock.list_foundation_models()
-        bedrock_models = response.get("modelSummaries", [])
+        response = bedrock.list_foundation_models(
+            byInferenceType="ON_DEMAND", byOutputModality="TEXT"
```
bigadsoleiman, line 59: Can we leverage enums instead of hardcoded values?

https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/shared/layers/python-sdk/python/genai_core/types.py

probably need 2 new enums for inference type and model life cycle status

bigadsoleiman (changes requested):

lib/shared/layers/python-sdk/python/genai_core/models.py
massi-ang: Would be nice if these enums were provided by boto3 :-)

bigadsoleiman (approved):

Issue #222, opened 2023-11-17
Feature request: Option to disable cross encoder models


ajaylamba-provar:
Currently cross encoder models are used to rank the search results but the models available need to be hosted on Sagemaker which increases cost significantly. Having an option to disable cross encoder models would be helpful while exploring the chatbot so that Sagemaker costs can be avoided.
With a toggle on UI or an option during deployment would be helpful.

azaylamba:
Work is in progress as part of https://github.com/aws-samples/aws-genai-llm-chatbot/pull/286

PR #286, opened 2023-12-23
Feature request: Option to disable cross encoder models

azaylamba:
Issue #222
*Description of changes:* Currently cross encoder models are used to rank the search results but the models available need to be hosted on Sagemaker which increases cost significantly. Having an option to disable cross encoder models would be helpful while exploring the chatbot so that Sagemaker costs can be avoided.

Added a config to enable/disable embeddings via Sagemaker which in turn derives cross encoding models. 
Persisted enableSagemakerModels config so that it can be used directly instead of relying on sagemakerModels length.


bigadsoleiman (approved):

spugachev:
@azaylamba We have migrated to AppSync. This PR has conflicts. Could you please fix this? And than we can merge

azaylamba:
@spugachev I couldn't find any conflicts in the PR. I have merged the main branch.

Note: I have not tested the changes after syncing with main due to some constraints with my AWS account. Any help in testing the changes is appreciated.

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -37,12 +38,13 @@ def query_workspace_aurora(
     if selected_model is None:
         raise genai_core.types.CommonError("Embeddings model not found")
 
-    cross_encoder_model = genai_core.cross_encoder.get_cross_encoder_model(
-        cross_encoder_model_provider, cross_encoder_model_name
-    )
+    if (config["rag"]["crossEncodingEnabled"]):
```
massi-ang, line 41: Move this part of the code below and have a single `if` statement

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -185,78 +187,69 @@ def query_workspace_aurora(
...
-            unique_items[i]["score"] = score
-            score_dict[unique_items[i]["chunk_id"]] = score
-
-    unique_items = sorted(unique_items, key=lambda x: x["score"], reverse=True)
-
-    for record in vector_search_records:
-        record["score"] = score_dict[record["chunk_id"]]
-    for record in keyword_search_records:
-        record["score"] = score_dict[record["chunk_id"]]
+    if (config["rag"]["crossEncodingEnabled"]):
```
massi-ang, line 190: It would be better that everything that is conditional to having enable crossEncodingModels is under a single `if` statement. This should be the one

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -185,78 +187,69 @@ def query_workspace_aurora(
...
-                )
-            else:
+            for i in range(len(unique_items)):
+                score = passage_scores[i]
+                unique_items[i]["score"] = score
+                score_dict[unique_items[i]["chunk_id"]] = score
+
+        unique_items = sorted(unique_items, key=lambda x: x["score"], reverse=True)
+
+    if (config["rag"]["crossEncodingEnabled"]):
```
massi-ang, line 205: remove this `if` statement

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -185,78 +187,69 @@ def query_workspace_aurora(
...
-            "items": convert_types(unique_items),
-            "vector_search_metric": metric,
-            "vector_search_items": convert_types(vector_search_records),
-            "keyword_search_items": convert_types(keyword_search_records),
-        }
-    else:
-        ret_items = list(filter(lambda val: val["score"] > threshold, unique_items))[
-            :limit
-        ]
-        if len(ret_items) < limit:
```
massi-ang, line 223: Why this code has been removed?

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -185,78 +187,69 @@ def query_workspace_aurora(
...
-                    reverse=True,
+                    unique_items, key=lambda x: x["vector_search_score"], reverse=True
                 )
                 ret_items = ret_items + (
                     list(
-                        filter(
-                            lambda val: (val["vector_search_score"] or -1) > 0.5,
-                            unique_items,
-                        )
+                        filter(lambda val: val["vector_search_score"] > 0.5, unique_items)
```
massi-ang, line 230: How is this change related to making cross encoders optional?

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
massi-ang, line 1: Regroup all the functionality related to cross encoders and have a single `if` statement. Check also for changes that have nothing to do with making cross-encoders optional.

lib/shared/layers/python-sdk/python/genai_core/opensearch/query.py
massi-ang, line 1: Regroup all the functionality related to cross encoders and have a single `if` statement. Check also for changes that have nothing to do with making cross-encoders optional.

massi-ang (changes requested):
On top of my comments on the changes, I would expect changes to the front end. If cross encoder models are not enabled the the menu should not be displayed. 
Also, if cross encoder models are not selected, it should not be possible to enable hybrid search on the workspaces.

azaylamba:
@massi-ang It seems syncing with main has caused some unwanted changes, I will work on this.

azaylamba:
@massi-ang I have addressed the review comments, please have a look.

cli/magic-create.ts
```diff
@@ -203,6 +203,24 @@ async function processCreateOptions(options: any): Promise<void> {
...
+      type: "confirm",
+      name: "enableEmbeddingModelsViaSagemaker",
+      message: "Do you want to enable embedding models via SageMaker?",
+      initial: options.enableEmbeddingModelsViaSagemaker || false,
+      skip(): boolean {
+        return !(this as any).state.answers.enableRag;
+      },
+    },
+    {
+      type: "confirm",
```
massi-ang, line 216: I do not see the point to have 2 different questions: if the users enables Embedding models via SM, then enabling Cross Encoders does not cost more. And it is not possible to enable Cross Encoders without enabling SM

cli/magic-create.ts
```diff
@@ -377,12 +399,24 @@ async function processCreateOptions(options: any): Promise<void> {
     models.defaultEmbedding = embeddingModels[0].name;
   }
 
-  config.rag.crossEncoderModels[0] = {
-    provider: "sagemaker",
-    name: "cross-encoder/ms-marco-MiniLM-L-12-v2",
-    default: true,
-  };
-  config.rag.embeddingsModels = embeddingModels;
+  if (answers.enableCrossEncoding && answers.enableSagemakerModels) {
```
massi-ang, line 402: if `enableSagemakerModels === true` you can deploy both embeddings and cross encoder models.

massi-ang (changes requested):
I do not see the point in having 2 settings (Cross Encoder / Embeddings) , since if you enable embeddings models via SM, you can get Cross Encoder for free.

massi-ang:
I think the configuration should be simple and meaningful for the user not the backend. You could think of an alternative naming for the parameter if you think that could create confusion, or better, you could  create a function called `is_cross_encoding_enabled` and use that in all your backend logic. In this way your code is self describing. In the function you can add a comment explaining why enabling embeddings is equivalent to enabling the cross encoder (with the current implementation). This approach provides an easy way to make the code evolve in the future.

azaylamba:
@massi-ang I have removed the prompt for `crossEncodingEnabled` and now it is being driven from the config `enableEmbeddingModelsViaSagemaker`. The reason I still kept `crossEncodingEnabled` in the config is because similar parameter (`crossEncodersEnabled`) is already being used in existing code.

azaylamba:
@massi-ang Would you be able to have a look at this again?

azaylamba:
@massi-ang Please let me know if more changes are required on this one.

toeteuf:
@massi-ang  I wanted to follow up on this PR submitted by @azaylamba  that is still pending review. Your feedback and approval are crucial for us, we also would like this feature. Regards

lib/rag-engines/index.ts
```diff
@@ -41,10 +41,7 @@ export class RagEngines extends Construct {
     const tables = new RagDynamoDBTables(this, "RagDynamoDBTables");
 
     let sageMakerRagModels: SageMakerRagModels | null = null;
-    if (
-      props.config.rag.engines.aurora.enabled ||
-      props.config.rag.engines.opensearch.enabled
-    ) {
+    if (props.config.llms.enableSagemakerModels) {
```
massi-ang, line 44: This should be checking `crossEncodingEnabled` and not `enableSageMakerModels`

lib/user-interface/react-app/src/pages/rag/create-workspace/aurora-form.tsx
```diff
@@ -112,18 +112,22 @@ function AuroraFooter(props: {
...
-          checked={props.data.hybridSearch}
-          onChange={props.onChange}
-        />
-        <CrossEncoderSelectorField
-          errors={props.errors}
-          submitting={props.submitting}
-          selectedModel={props.data.crossEncoderModel}
-          onChange={props.onChange}
-        />
+        {props.data.crossEncodingEnabled && (
```
massi-ang, line 115: Instead of hiding the control, disable them if `!crossEncodingEnabled`

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-aurora.tsx
```diff
@@ -30,11 +30,12 @@ const metrics = [
 const defaults: AuroraWorkspaceCreateInput = {
   name: "",
   embeddingsModel: null,
+  crossEncodingEnabled: false,
```
massi-ang, line 33: the default should reflect the value of the configuration

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-aurora.tsx
```diff
@@ -30,11 +30,12 @@ const metrics = [
...
   crossEncoderModel: null,
   languages: [{ value: "english", label: "English" }],
   metric: metrics[0].value,
   index: true,
-  hybridSearch: true,
+  hybridSearch: false,
```
massi-ang, line 38: should be `true` if cross encoders have been enabled in the configuration

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-opensearch.tsx
```diff
@@ -15,9 +15,10 @@ const nameRegex = /^[\w+_-]+$/;
 const defaults: OpenSearchWorkspaceCreateInput = {
   name: "",
   embeddingsModel: null,
+  crossEncodingEnabled: false,
```
massi-ang, line 18: default value should reflect the value from the configuration

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-opensearch.tsx
```diff
@@ -15,9 +15,10 @@ const nameRegex = /^[\w+_-]+$/;
...
   crossEncoderModel: null,
   languages: [{ value: "english", label: "English" }],
-  hybridSearch: true,
+  hybridSearch: false,
```
massi-ang, line 21: should be true if cross encoders have been enabled in the configuration

lib/user-interface/react-app/src/pages/rag/create-workspace/opensearch-form.tsx
```diff
@@ -70,18 +70,22 @@ function OpenSearchFooter(props: {
...
-          checked={props.data.hybridSearch}
-          onChange={props.onChange}
-        />
-        <CrossEncoderSelectorField
-          errors={props.errors}
-          submitting={props.submitting}
-          selectedModel={props.data.crossEncoderModel}
-          onChange={props.onChange}
-        />
+        {props.data.crossEncodingEnabled && (
```
massi-ang, line 73: disable the controls if `!crossEncodingEnabled` instead of hiding them

cli/magic-config.ts
```diff
@@ -546,6 +550,15 @@ async function processCreateOptions(options: any): Promise<void> {
       message: "Do you want to enable RAG",
       initial: options.enableRag || false,
     },
+    {
+      type: "confirm",
+      name: "enableEmbeddingModelsViaSagemaker",
+      message: "Do you want to enable embedding models via SageMaker?",
```
massi-ang, line 556: For now I suggest to change it to "Do you want to enable embedding and cross-encoder models via SageMaker?" to clarify the consequences of the choice to the users.

When the  PR to enable cohere cross encoders via Cohere API will be merged, we will  add a separate question about enabling cross-encoders for hybrid search.

cli/magic-config.ts
```diff
@@ -546,6 +550,15 @@ async function processCreateOptions(options: any): Promise<void> {
...
+      initial: options.enableEmbeddingModelsViaSagemaker || false,
```
massi-ang, line 557: this does not seem to be working, and initial value is always `false` in the UI

cli/magic-config.ts
```diff
@@ -755,28 +770,44 @@ async function processCreateOptions(options: any): Promise<void> {
...
+      name: "cross-encoder/ms-marco-MiniLM-L-12-v2",
+      default: true,
+    };
+  } else {
+    config.rag.crossEncoderModels[0] = {
+      provider: "None",
+      name: "None",
+      default: true,
+    };
+  }
```
massi-ang, line 794: This configuration should not be conditional on the fact that embedding models have been enabled via SM. Always set:
```
config.rag.crossEncoderModels[0] = {
      provider: "sagemaker",
      name: "cross-encoder/ms-marco-MiniLM-L-12-v2",
      default: true,
    };
```

cli/magic-config.ts
```diff
@@ -755,28 +770,44 @@ async function processCreateOptions(options: any): Promise<void> {
...
+  if (!config.rag.enableEmbeddingModelsViaSagemaker) {
+    config.rag.embeddingsModels = embeddingModels.filter(model => model.provider !== "sagemaker");
```
massi-ang, line 796: This logic should also be applied to the list of models shown in the UI when selecting the default embedding model.

massi-ang:
Hi @azaylamba, 
Please look at my comments and fix accordingly. Thanks.

massi-ang (changes requested):
Good work but few fixes left

cli/magic-config.ts
azaylamba: We should make it clear that the cross encoding is not enabled. By setting the model as None, user would get to know that there is no cross encoder model. I will change the if condition to check for `config.rag.crossEncodingEnabled`

lib/rag-engines/index.ts
azaylamba: ok, but won't that be confusing that crossEncodingEnabled is driving the Sagemaker models instead of the config `props.config.llms.enableSagemakerModels` which is specific for sagemaker models?

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-opensearch.tsx
azaylamba: Here we are just defining the constant, the value from the configuration is read below at line 38.

lib/user-interface/react-app/src/pages/rag/create-workspace/create-workspace-opensearch.tsx
azaylamba: Here we are just defining the constant, the value from the configuration is read below at line 38.

cli/magic-config.ts
azaylamba: corrected now

azaylamba:
@massi-ang Addressed the review comments, please have a look.

cli/magic-config.ts
massi-ang: Please correct as I mentioned, this configuration just reflects which are the cross-encoder models that the solution can use and is independent from the setting `config.rag.crossEncodingEnabled`. The behaviour should be the same as for the embedding models. If we disable sagemaker embedding models we are still leaving the configuration for the embbeddings.

cli/magic-config.ts
```diff
@@ -1108,7 +1107,9 @@ async function processCreateOptions(options: any): Promise<void> {
     };
   }
```
massi-ang, line 1108: For now remove the `if` and always add the cross encoder model to the configuration. The `crossEncodingEnabled` setting is already defining if cross encoders are enabled or not. This should be consistent on how we handle embedding models. A possible alternative in both cases would be to put an empty list in case they are disabled.

lib/user-interface/react-app/src/common/types.ts
```diff
@@ -25,6 +25,7 @@ export interface AppConfig {
     rag_enabled: boolean;
     cross_encoders_enabled: boolean;
     sagemaker_embeddings_enabled: boolean;
+    enable_embedding_models_via_sagemaker: boolean;
```
massi-ang, line 28: What is the difference between the two?

massi-ang (changes requested):

lib/user-interface/react-app/src/common/types.ts
azaylamba: Thanks for pointing out, I misinterpreted the other config. Removed it now.

lib/rag-engines/index.ts
massi-ang: You are right and `props.config.llms.enableSagemakerModels` is better

lib/rag-engines/index.ts
```diff
@@ -42,10 +42,7 @@ export class RagEngines extends Construct {
     const tables = new RagDynamoDBTables(this, "RagDynamoDBTables");
 
     let sageMakerRagModels: SageMakerRagModels | null = null;
-    if (
-      props.config.rag.engines.aurora.enabled ||
-      props.config.rag.engines.opensearch.enabled
-    ) {
+    if (props.config.rag.crossEncodingEnabled) {
```
massi-ang, line 45: You were right, `props.config.llms.enableSagemakerModels` is better.

massi-ang (changes requested):
One last change and it seems to be all good.

lib/rag-engines/index.ts
azaylamba: Updated the condition.

azaylamba:
@massi-ang Addressed the last review comment, please have a look.

toeteuf:
@massi-ang I think @azaylamba  commit the last change requested... Could you please have a look?

charles-marion:
Hi @azaylamba , @toeteuf ,

Apologies for the delay. 

I am making changes based on your PR to fix/set the list of models in the config instead of adding new properties (and reviewing your change at the same time).

I will most likely create a new PR with these changes and follow up until merged (and mention you are the original author @azaylamba ). I will also verify the unit tests/format is passing verifications.

Please tell me if you have any concern.

azaylamba:
> Hi @azaylamba , @toeteuf ,
> 
> Apologies for the delay.
> 
> I am making changes based on your PR to fix/set the list of models in the config instead of adding new properties (and reviewing your change at the same time).
> 
> I will most likely create a new PR with these changes and follow up until merged (and mention you are the original author @azaylamba ). I will also verify the unit tests/format is passing verifications.
> 
> Please tell me if you have any concern.

I don't have any objection, please proceed with the changes.

charles-marion:
Closed in favor of #588

Issue #151, opened 2023-10-27
Enhancement: add answer feeedback ratings


massi-ang:
While testing a new chatbot it is important to be able to gather information about the quality of the answers. Since quality might mean different things for different use cases, the "quality ratings" should be configurable per chatbot. 

This enhancement is about adding the possibility to define up to 5 rating criteria and the associated values. Rating criteria can then be associated with a chatbot and displayed in the UI at the bottom of each answer provided by the system. 
As the user changes the rating, they are stored in the chat history.

A chatbot administrator would then be able to get a report about all the answer rating and their associated metadata.

azaylamba:
Changes are done and are out for review under https://github.com/aws-samples/aws-genai-llm-chatbot/pull/287

azaylamba:
https://github.com/aws-samples/aws-genai-llm-chatbot/pull/287 has been merged, this issue can be closed.

PR #287, opened 2023-12-24
Enhancement: Add user feedback for responses

azaylamba:
***Issue*** #151 
***Description of changes:*** Added basic feedback mechanism for responses generated by the chatbot. 
As part of this change, we have stored the user feedback in S3 bucket.

We are storing the feedback in the following format:
{
"feedbackId": "<feedbackId>",
"sessionId": "<sessionId>",
"userId": "<userId>",
"key": "<key>",
"prompt": "<question>",
"completion": "<generated text>",
"model": "<model name>",
"feedback": "1/0",
"createdAt": "<timestamp>"
}

This feedback can be used by AWS Athena/Glue etc to do analysis and also to train the model if required. We are keeping one object per feedback in S3 'STANDARD_IA' class.


bigadsoleiman:
Hi @azaylamba thanks for your submission. I really like this feature as it opens up to offline analyses and more importantly fine tuning of models where possible and I think this should be the end goal.

Though, I think this should be stored in a form of dataset that then can be used for fine-tuning. See Bedrock's [prepare for finetuning docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-customization-prepare.html#model-customization-prepare-finetuning) which requires dataset in the followin format

```
{"prompt": "<prompt text>", "completion": "<expected generated text>"}
```

now in this case we can bring on more fields such as 

```
{
   "prompt": "<prompt text>", 
   "completion": "<expected generated text>", 
   "model" ..., 
   "feedkback": 0/1,
   "timestamp":  ..., 
   ...any other data of interest
}
```
and stored in S3 maybe with a Infrequent Access storage class.

The advantages would be:
- Cheaper option
- Admins or systems can leverage Athena to query this feedbacks
- Glue can be used to re-format this feedbacks for model-specific finetuning

You can even store an object per feedback, Athena, Glue/EMR will do the heavy-lifting to merge them when needed.

azaylamba:
@bigadsoleiman I have modified the code to store the user feedback in S3 bucket as per the suggested format. Please have a look.

lib/shared/layers/python-sdk/python/genai_core/user_feedback.py
```diff
@@ -0,0 +1,51 @@
...
+        "feedback": feedback,
+        "createdAt": timestamp
+    }
+    
+    response = s3_client.put_object(
+        Bucket=USER_FEEDBACK_BUCKET_NAME,
+        Key=feedbackId,
+        Body=json.dumps(item),
+        ContentType="application/json",
+        StorageClass='STANDARD_IA',
```
bigadsoleiman, line 43: Let's handle storage class in CDK instead.

something along these line should ensure that objects are transitioned to the right class and it's easier to handle more complex rules in future

```
    const userFeedbackBucket = new s3.Bucket(this, 'UserFeedbackBucket', {
       ...
      lifecycleRules: [{
        transitions: [{
          storageClass: s3.StorageClass.INFREQUENT_ACCESS,
          transitionAfter: cdk.Duration.days(0), // Transition immediately after creation
        }],
      }],
    });
  }
```

lib/chatbot-api/chatbot-s3-buckets/index.ts
```diff
@@ -39,7 +40,31 @@ export class ChatBotS3Buckets extends Construct {
       ],
     });
 
+    const userFeedbackBucket = new s3.Bucket(this, "UserFeedbackBucket", {
```
bigadsoleiman, line 43: let's set the lifecycle rules here for simplicity and having them handled in one place (instead of object creation)

```
    const userFeedbackBucket = new s3.Bucket(this, 'UserFeedbackBucket', {
       ...
      lifecycleRules: [{
        transitions: [{
          storageClass: s3.StorageClass.INFREQUENT_ACCESS,
          transitionAfter: cdk.Duration.days(0), // Transition immediately after creation
        }],
      }],
    });
  }
 ```

File: lib/chatbot-api/chatbot-s3-buckets/index.ts [hunk 3]
```diff
@@ -39,7 +40,31 @@ export class ChatBotS3Buckets extends Construct {
...
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      transferAcceleration: true,
+      enforceSSL: true,
+      serverAccessLogsBucket: logsBucket,
+      cors: [
```
bigadsoleiman, line 50: Why CORS are enabled here? We are not dealing with objects directly from web clients

lib/chatbot-api/chatbot-s3-buckets/index.ts
```diff
@@ -39,7 +40,31 @@ export class ChatBotS3Buckets extends Construct {
       ],
     });
 
+    const userFeedbackBucket = new s3.Bucket(this, "UserFeedbackBucket", {
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: true,
+      transferAcceleration: true,
```
bigadsoleiman, line 47: Don't think we need transfer acceleration here right?

bigadsoleiman (changes requested):
Thanks!! going in right direction.

One thing I would change is the value of the feedback, I would avoid strings such as `thumbsUp/thumbsDown` as it's easier in a context of EMR, Athena to deal with numbers especially when it comes to aggregations so I'd rather use `0/1`

Unless there is a specific reason I'm missing for using `thumbsUp/thumbsDown`, let me know.

lib/chatbot-api/functions/api-handler/routes/user_feedback.py
```diff
@@ -0,0 +1,37 @@
...
+logger = Logger()
+
+
+class CreateUserFeedbackRequest(BaseModel):
+    sessionId: str
+    key: str
+    feedback: str
+    prompt: str
+    completion: str
+    model: str
```
bigadsoleiman, line 19: let's include the `userId` too

bigadsoleiman (changes requested):

lib/shared/layers/python-sdk/python/genai_core/user_feedback.py
azaylamba: The minimum allowed value of transitionAfter days is 30. The recommended approach to directly store the objects in STANDARD_IA is setting the storage class at object level.

bigadsoleiman (approved):

azaylamba:
thanks for your inputs @bigadsoleiman

Issue #289, opened 2023-12-27
All the sessions are not displayed in the history


ajaylamba-provar:
I am facing an issue where all the chat sessions of a user are not loaded in the history session. When I am starting a new session, the data is being stored in the DynamoDB table `SessionsTable`. I can explore the items from DynamoDB console manually. 
But all the items are not loaded on the UI. The UI returns only 41 sessions for that user while the dynamo db table has 179 records for that user.

The following code in `sessions.tsx` class returns only 41 records consistently for that user. While for other users, the issue is not reproducible.

```
  const getSessions = useCallback(async () => {
    if (!appContext) return;

    const apiClient = new ApiClient(appContext);
    const result = await apiClient.sessions.getSessions();
    if (ResultValue.ok(result)) {
      setSessions(result.data);
    }
  }, [appContext]);
```

PR #290, opened 2023-12-27
All the sessions are not displayed in the history

azaylamba:
*Issue #289:* For the users who have long conversation histories and a lot of items in sessions table, all the records were not displayed on the UI.

*Description of changes:* DynamoDB query operation was not returning all the records because the total size of records was more than 1MB and in such cases, DynamoDB query operation does not handle pagination automatically.

As a fix, added code to handle the pagination based on the 'LastEvaluatedKey' returned by DynamoDB.


azaylamba:
@massi-ang could you please have a look?

PR #294, opened 2024-01-04
Fix: PGVector upgrade and hnsw indexing

rcgeorge:
*Issue #, if available:*
162, 235

*Description of changes:*
This change updates the Aurora Postgres engine version from 15.3 to 15.5 to support the pgvector v0.5.1 extension. The indexing has been changed from ivfflat to hnsw.


bigadsoleiman:
Hi @rcgeorge thanks for your submission!

1) What would be rationale for switching to `hnsw` across the board? or it's for your specific use case?
2) I'm asking as it would probably be better to allow the freedom to select the index having `ivfflat` as default with the ability to switch to `hnsw`

spugachev:
It seems that this would break existing installations. I would prefer to add hnsw as the new default option but keep ivfflat for existing ivfflat users. Manu users have this deployed. So please don't rename columns. Another option would be to write a data migration script.

Issue #306, opened 2024-01-16
ability disable use of chat history context in following answers


Rob-Powell:
Sometimes users don't want to enable contextual awareness of the conversation and each question within the same window should be treated as independent.

It would be useful to be able to disable chat history context usage in formulating a reply, without affecting the logging of the chats.

bigadsoleiman:
After review the best path forward would be to make it easier to create a new session instead of allowing no-history in conversation.

Closing in favour of https://github.com/aws-samples/aws-genai-llm-chatbot/issues/283

Issue #283, opened 2023-12-20
Feature: introduce a more accessible way to start a new conversation in the playgrounds


massi-ang:
Currently the chatbot allows to start new conversation by opening the right side panel and clicking on "New Session". This is fine, but having a mechanism (like a button) in the main UI would make it easier to start new conversations.

Rob-Powell:
I agree this is a good feature to have but not a replacement for #306 clicking this button every time if you know each line entered is to be unique is simply redundant. @bigadsoleiman

PR #307, opened 2024-01-16
feat: enable/disable use of history in chat context

Rob-Powell:
*Issue #, if available:* #306

*Description of changes:*
Allows customers to disable chat history context usage without affecting the recording of chat history for logging purposes. This setting is implemented on a per workspace level.

The updated UI elements on the create workspace workflow further explain the setting to the users and it is enabled by default to retain existing functionality.

I have tested the changes.


bigadsoleiman:
After review the best path forward would be to make it easier to create a new session instead of allowing no-history in conversation.

Closing in favour of https://github.com/aws-samples/aws-genai-llm-chatbot/issues/283

PR #308, opened 2024-01-17
feat: ability to run amplify command to generate code files

### LINKED ISSUE #278: Improve Amplify Code Generation
Introduced in PR #269, The react app is leveraging GraphQL for the API. As part of this, Amplify codegen will create the Mutations, Query & Subscription TypeScript Objects, as well as the API object. 

With the changes, the generated files are included, but there's not an easy-to-update way to change the generated code without deploying the back-end first. 

It would be preferred if an `npm run` command or something similar could be added to improve the developer experience when updating the API

Rob-Powell:
*Issue #, if available:* #278

*Description of changes:*
This resolves #278 by implemented a dev command `npm run gen` which runs the amplify cli to generate files based off schema.graphql reducing developer burden with appsync solution.

https://docs.amplify.aws/javascript/build-a-backend/graphqlapi/client-code-generation/


bigadsoleiman (approved):

.graphqlconfig.yml
```diff
@@ -0,0 +1,20 @@
+projects:
+  test:
```
massi-ang, line 2: maybe change the name of the project to `chatbot`

massi-ang:
Thanks for this contribution. Please remove the `schema.json` file from the `lib/user-interface/react-app` since it is not used and will never be refreshed.

massi-ang (changes requested):
This change makes the `lib/user-interface/react-app/schema.json` redundant. Please remove.

.graphqlconfig.yml
Rob-Powell: updated

massi-ang (approved):

Issue #312, opened 2024-01-19
support for mistral instruct 0.2 and mixtrail instruct models


Rob-Powell:
support for mistral instruct 0.2 and mixtrail instruct models and sagemaker hosted models.

These are still quite important for countries without access to bedrock and have strong data sovereignty requirements.

bigadsoleiman:
In addition
- Update documentation for adding new model

PR #313, opened 2024-01-19
feat: adding support for mixtral 0.1 and mistral instruct 0.2

Rob-Powell:
*Issue #, if available:* 
#312

*Description of changes:*
Adding support for Mixtral instruct 0.1 and Mistral instruct 0.2
added updated container image references which are required to support the newer models
tested the changes

### DIFF
--- a/lib/sagemaker-model/container-images.ts
+++ b/lib/sagemaker-model/container-images.ts
@@ -24,6 +24,8 @@ export class ContainerImages {
     "huggingface-pytorch-tgi-inference:2.0.1-tgi1.0.3-gpu-py39-cu118-ubuntu20.04";
   static readonly HF_PYTORCH_LLM_TGI_INFERENCE_1_1_0 =
     "huggingface-pytorch-tgi-inference:2.0.1-tgi1.1.0-gpu-py39-cu118-ubuntu20.04";
+  static readonly HF_PYTORCH_LLM_TGI_INFERENCE_1_3_3 =
+    "huggingface-pytorch-tgi-inference:2.1.1-tgi1.3.3-gpu-py310-cu121-ubuntu20.04";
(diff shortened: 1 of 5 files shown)


lib/sagemaker-model/container-images.ts
```diff
@@ -24,8 +24,10 @@ export class ContainerImages {
     "huggingface-pytorch-tgi-inference:2.0.1-tgi1.0.3-gpu-py39-cu118-ubuntu20.04";
   static readonly HF_PYTORCH_LLM_TGI_INFERENCE_1_1_0 =
     "huggingface-pytorch-tgi-inference:2.0.1-tgi1.1.0-gpu-py39-cu118-ubuntu20.04";
+  static readonly HF_PYTORCH_LLM_TGI_INFERENCE_1_3_3 =
+    "huggingface-pytorch-tgi-inference:2.1.1-tgi1.3.3-gpu-py310-cu121-ubuntu20.04";
   static readonly HF_PYTORCH_LLM_TGI_INFERENCE_LATEST =
-    ContainerImages.HF_PYTORCH_LLM_TGI_INFERENCE_1_1_0;
+    ContainerImages.HF_PYTORCH_LLM_TGI_INFERENCE_1_3_3;
```
bigadsoleiman, line 30: Let's not update/use LATEST as it might impact other models running on previous TGI version unexpectedly. To be fair we should remove LATEST from this but it's out of scope of this PR. 

So please just change mi(s|x)tral models to point to the explicit version instead of LATEST.

bigadsoleiman (changes requested):

lib/sagemaker-model/container-images.ts
Rob-Powell: have done this and squashed commits

bigadsoleiman (approved):

PR #320, opened 2024-01-19
Attempt local asset bundling before using Docker.

dfkunstler:
*Issue #, if available:* N/A

*Description of changes:*
Implemented local asset bundling, which will be used instead of Docker when possible. Local bundling is preferred over DIND for performance and security reasons. This is the same approach already used to [bundle node.js dependencies in the UI construct](https://vscode.dev/github/dfkunstler/aws-genai-llm-chatbot/blob/main/lib/user-interface/index.ts#L203).

**Problem**: Docker-in-Docker execution is often blocked in regulated environments. CDK uses Docker for S3 asset bundling by default. `cdk synth` fails if Docker is not available, making it impossible to build or deploy the solution. 

**Solution**: This implementation of local bundling should succeed if `zip` and `python3.11` (with `pip`) are available on the local path. If not, CDK falls back to Docker, so worst case we're back to the stated problem. Note that the required version of python is determined by the Lambda runtime specified in [lib/shared/index.ts](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/shared/index.ts#L12). Hard-coding `python3` instead would probably make this work for more people, but matching the Lambda runtime version seemed a safer choice. I'll leave it to reviewers to make the final call.

Due to limits in CDK (and perhaps my knowledge of Javascript), I had to add similar code to two different classes. Currently, the CDK [ILocalBundling.tryBundle(...)](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.ILocalBundling.html) interface definition does not include the `path` to be bundled. An inline anonymous class implementation is the only solution I could get to work. 

I've tested this on multiple versions of Amazon Linux and on Mac OS, but nowhere else. 

Bundling discussion in CDK docs:
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3_assets-readme.html#asset-bundling

            

### DIFF
--- a/lib/layer/index.ts
+++ b/lib/layer/index.ts
@@ -1,6 +1,8 @@
 import * as cdk from "aws-cdk-lib";
 import * as lambda from "aws-cdk-lib/aws-lambda";
 import * as s3assets from "aws-cdk-lib/aws-s3-assets";
+import * as lpath from "path";
+import { execSync } from "child_process";
 import { Construct } from "constructs";
 
 interface LayerProps {
@@ -26,6 +28,47 @@ export class Layer extends Construct {
     const layerAsset = new s3assets.Asset(this, "LayerAsset", {
       path,
       bundling: {
+        local: {
+          /* implements a local method of bundling that does not depend on Docker. Local
+          bundling is preferred over DIND for performance and security reasons.
+          see https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.ILocalBundling.html and 
+          https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3_assets-readme.html#asset-bundling */    
+          tryBundle(outputDir: string, options: cdk.BundlingOptions) {
+            let canRunLocal = false;
+            let python = props.runtime.name;
+
+            try {
+              // check if pip is available locally
+              const testCommand = `${python} -m pip -V`
+              console.log(`Checking for pip: ${testCommand}`)
+              // without the stdio arg no output is printed to console
+              execSync(testCommand, { stdio: 'inherit' });
+              // no exception means command executed successfully
+              canRunLocal = true;
+            } catch {
+              // execSync throws Error in case return value of child process is non-zero.
+              // Actual output should be printed to the console.
+              console.warn(`Unable to do local bundling! ${python} with pip must be on path.`);
+            }
+
+            if (canRunLocal) {
+              const command = `${python} -m pip install -r ${lpath.posix.join(path, "requirements.txt")} -t ${outputDir} ${autoUpgrade ? '-U' : ''}`;
+              try {
+                console.debug(`Local bundling: ${command}`);
+                // this is where the work gets done
+                execSync(command, { stdio: 'inherit' });
+                return true;
+              } catch (ex) {
+                // execSync throws Error in case return value of child process
+                // is non-zero. It'll be printed to the console because of the 
+                // stdio argument.                             
+                console.log(`Local bundling attempt failed: ${ex}`)
+              }
+            }
+            // if we get here then Docker will be used as configured below
+            return false;
+          }
+        },
         image: runtime.bundlingImage,
         platform: architecture.dockerPlatform,
         command: [
--- a/lib/shared/shared-asset-bundler.ts
+++ b/lib/shared/shared-asset-bundler.ts
@@ -3,6 +3,7 @@ import {
   BundlingOutput,
   DockerImage,
   aws_s3_assets,
+  BundlingOptions
 } from "aws-cdk-lib";
 import { Code, S3Code } from "aws-cdk-lib/aws-lambda";
 import { Asset } from "aws-cdk-lib/aws-s3-assets";
@@ -10,6 +11,7 @@ import { md5hash } from "aws-cdk-lib/core/lib/helpers-internal";
 import { Construct } from "constructs";
 import * as path from "path";
 import * as fs from "fs";
+import { execSync } from "child_process";
 
 function calculateHash(paths: string[]): string {
   return paths.reduce((mh, p) => {
@@ -33,6 +35,8 @@ function calculateHash(paths: string[]): string {
 export class SharedAssetBundler extends Construct {
   private readonly sharedAssets: string[];
   private readonly WORKING_PATH = "/asset-input/";
+  private readonly container_image: DockerImage;
+  private useLocalBundler: boolean = false;
   /**
    * Instantiate a new SharedAssetBundler. You then invoke `bundleWithAsset(pathToAsset)` to
    * bundle your asset code with the common code.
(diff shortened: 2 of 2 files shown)


lib/shared/shared-asset-bundler.ts
```diff
@@ -47,17 +51,75 @@ export class SharedAssetBundler extends Construct {
   constructor(scope: Construct, id: string, sharedAssets: string[]) {
     super(scope, id);
     this.sharedAssets = sharedAssets;
+    // Check if we can do local bundling
+    if (!this.localBundlerTest()) {
+      // if not, then build Apline from local definition
```
massi-ang, line 56: Alpine

lib/shared/shared-asset-bundler.ts
```diff
@@ -47,17 +51,75 @@ export class SharedAssetBundler extends Construct {
...
+      this.container_image = DockerImage.fromBuild(path.posix.join(__dirname, "alpine-zip"));
```
massi-ang, line 57: shouldn't the `containerImage` be a singleton to avoid rebuilding the container for every use of a SharedAssetBundler construct?

lib/shared/shared-asset-bundler.ts
```diff
@@ -33,6 +35,8 @@ function calculateHash(paths: string[]): string {
 export class SharedAssetBundler extends Construct {
   private readonly sharedAssets: string[];
   private readonly WORKING_PATH = "/asset-input/";
+  private readonly container_image: DockerImage;
```
massi-ang, line 38: camelCase

lib/layer/index.ts
massi-ang, line 1: I would not implement local bundling for layers, especially because we require ML libraries that need to be compiled for the target architecture and local bundling will use the architecture of the local machine and not that of the Lambda compute configuration.

massi-ang (commented):
I think this is good change. Please see the few comments I have regarding naming and more importantly on using a singleton implementation to avoid trying to rebuild the container at every use of the SharedAssetBundler.

I would remove local bundling for the Lambda layers since it will create issues with dependencies that are platform specific, such as many ML libraries.

lib/layer/index.ts
dfkunstler: Good catch - thanks. Instead of skipping it altogether, what if we check target architecture vs local machine architecture? If they match, then we can try local bundling; if not, then fall back to Docker. Clearly there is no point building libs for the wrong platform, but if we skip it altogether then this PR doesn't solve the stated problem in any case. Since architecture is [currently set here](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/shared/index.ts#L15) to `X86_64`, i assume it'd still be useful for many people.

lib/shared/shared-asset-bundler.ts
dfkunstler: makes sense to me, but the [current implementation](https://github.com/aws-samples/aws-genai-llm-chatbot/pull/320/commits/d7b1c6a4b723a5c71112219fc804c5c39fc87cf0#diff-36fe271369c32e3bbc80d20fe50ebe4be930a8015edcb6c946fb8edc6e62e4c1L60) builds a new image with each invocation of method `bundleWithAsset(...)`, so I thought perhaps a Singleton was problematic for some reason? I'm not sure why it would be a problem, but I stopped short of Singleton because this change improved build time and my knowledge of code base is limited.

lib/shared/shared-asset-bundler.ts
dfkunstler: fixed

lib/shared/shared-asset-bundler.ts
dfkunstler: fixed

lib/shared/shared-asset-bundler.ts
massi-ang: The current implementation should be improved :-)

lib/layer/index.ts
massi-ang: Makes sense, but I am not sure this is worth the effort. Most people using this will use Macs which are using ARM and therefore will not be able to benefit from this solution. The other most common way to deploy is via Cloud9, where using docker is not an issue. Basically this will benefit only Windows users in the end.

lib/shared/shared-asset-bundler.ts
dfkunstler: done

lib/layer/index.ts
dfkunstler: Windows users benefit as a side-effect, but I'm not so concerned with individual/personal laptops. Lots of security-conscious organizations gate deployments through their managed dev workstations or build/CI systems, which enforce more strict controls than laptops. Especially in container-based build systems, DIND execution is often (i.e., should be) prohibited because it must run in privileged mode. Even on dev workstations, which tend to be more permissive, Docker may or may not be available. 

IMO we should ideally eliminate all Docker dependencies from the code and instead use alternatives like local bundling. Where that's impractical, we should externalize container builds and publish artifacts that this code consumes. I acknowledge the additional overhead and complexity as non-trivial. However, this is a powerful and valuable bit of software, and I think it's a worthwhile investment to increase adoption. Hopefully this PR is a first step in that direction.

Note that i've updated the PR to include your requested changes for review. however, i haven't finished verifying the latest commit yet, so if you're inclined to accept these changes, please don't merge until I confirm.

thanks for your time and the great software!

dfkunstler: ok - from my perspective we're all set to merge.

dfkunstler: @massi-ang friendly reminder that this is ready for your final review

lib/shared/shared-asset-bundler.ts
massi-ang, line 1: This code still generates a new DockerImage construct every time you instantiate the `SharedAssetBundler` construct.

Consider this pattern:

1. add:

```
class BuildImageProvider extends Construct {
  public static getOrCreate(scope: Construct): BuildImageProvider {
    const stack = Stack.of(scope);
    const id = "SharedAssetBuildImageProvider";
    const x =
      (Node.of(stack).tryFindChild(id) as BuildImageProvider) ||
      new BuildImageProvider(stack, id);
    return x;
  }

  readonly image: DockerImage;

  constructor(scope: Construct, id: string) {
    super(scope, id);

    const buildImage = DockerImage.fromBuild(
      path.posix.join(__dirname, "alpine-zip")
    );

    this.image = buildImage;
  }
}
```

2: use it via 
```
bundling: {
          image: BuildImageProvider.getOrCreate(this).image,
```

massi-ang (changes requested):
Nice work. Still need to fix the issue with the bundling image.

lib/shared/shared-asset-bundler.ts
dfkunstler: @massi-ang sorry but i'm not sure i understand; my most recent commit uses a [static init block starting on line 47](https://github.com/aws-samples/aws-genai-llm-chatbot/pull/320/files#diff-36fe271369c32e3bbc80d20fe50ebe4be930a8015edcb6c946fb8edc6e62e4c1R47). this morning i got a code review from a colleague who agreed that it should not execute again on instantiations of `SharedAssetBundler`. 

For sanity checking, i then modified my local copy to show me when and in what context the static block executes (using reflection and stack dumps). my testing was far from exhaustive, but i did try a couple of different scenarios. the static block executes only when Node compiles the class on first ref; it does not execute again on instantiation.

maybe GitHub was showing you the prior commit? i'm far from an expert on Javascript so it's certainly possible my implementation and test method are wrong. but before i make more changes, could you please confirm? thanks for taking the time to explain.

massi-ang: I missed the use of `static`. This solution works as well.

massi-ang (dismissed):

massi-ang:
Hi, I am getting an issue with the code, and need to investigate the underlying reason

dfkunstler:
> Hi, I am getting an issue with the code, and need to investigate the underlying reason. 

Coincidentally, i've just tried a clean build of 4.0.2 and am getting an error related to [this import in the react app](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/user-interface/react-app/src/app.tsx#L19):
`import * as InfraConfig from '../../../../bin/config.json';`

Build output makes it seem like a bundling error, but i think the bundling error is a consequence of the compilation failure. Is this the same issue you're seeing? If so, keep in mind that the local bundling used in the UI construct was implemented previously and is not related to this PR.

massi-ang:
> > Hi, I am getting an issue with the code, and need to investigate the underlying reason.
> 
> Coincidentally, i've just tried a clean build of 4.0.2 and am getting an error related to [this import in the react app](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/user-interface/react-app/src/app.tsx#L19): `import * as InfraConfig from '../../../../bin/config.json';`
> 
> Build output makes it seem like a bundling error, but i think the bundling error is a consequence of the compilation failure. Is this the same issue you're seeing? If so, keep in mind that the local bundling used in the UI construct was implemented previously and is not related to this PR.

This is a different issue. Run `npm run build && npm run create` to create a new `config.json` and then try again.

dfkunstler:
> > > Hi, I am getting an issue with the code, and need to investigate the underlying reason.
> > 
> > 
> > Coincidentally, i've just tried a clean build of 4.0.2 and am getting an error related to [this import in the react app](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/user-interface/react-app/src/app.tsx#L19): `import * as InfraConfig from '../../../../bin/config.json';`
> > Build output makes it seem like a bundling error, but i think the bundling error is a consequence of the compilation failure. Is this the same issue you're seeing? If so, keep in mind that the local bundling used in the UI construct was implemented previously and is not related to this PR.
> 
> This might be an unrelated issue. Try running `npm run build && npm run create` to create a new `config.json` and then try again.

i'm confused about the decision to handle import that way; will open a separate issue to ask about it.

massi-ang:
The solution with the `static` class member does not work for me. The only solution that seems to work is to use this pattern:

```ts
class BuildImageProvider extends Construct {
  public static getOrCreate(scope: Construct): DockerImage {
    const stack = Stack.of(scope);
    const id = "build-image-provider";
    const x =
      (Node.of(stack).tryFindChild(id) as BuildImageProvider) ||
      new BuildImageProvider(stack, id);
    return x.buildImage;
  }

  private readonly buildImage: DockerImage;

  constructor(scope: Construct, id: string) {
    super(scope, id);

    this.buildImage = DockerImage.fromBuild(
      path.posix.join(__dirname, "alpine-zip")
    );
  }
}
```

and then retrieve the image builder with

```ts
bundling: {
          image: BuildImageProvider.getOrCreate(this),
```

dfkunstler:
@massi-ang any news on this?

PR #329, opened 2024-01-23
bug: docs are using old architecture diagram without appsync

Rob-Powell:
*Issue #, if available:*

*Description of changes:*
When moving to vitepress docs updated image with appsync commit was replaced with the original api gateway diagram, this corrects that with the proper image.


massi-ang (commented):
The resolution of the 2 files is not the same. Any issue with that?

Rob-Powell:
> The resolution of the 2 files is not the same. Any issue with that?

I just put back the same image size you originally committed with your changes @massi-ang. the docs site simply references the image, the new size won't cause any issues, aside from it being slightly smaller.

bigadsoleiman (approved):

PR #359, opened 2024-02-05
feat(embeddings): adding Cohere and Titan Multimodal

massi-ang:
*Issue #, if available:*
Resolves #218 

*Description of changes:*
Added support for Cohere Embeddings and Titan Multimodal.

- Titan multimodal can only be used for text embeddings
- Cohere Embeddings are created using `search_document` setting as suggested by Cohere documentation for multi-purpose embedding generation (document embeddings and query embeddings)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #218, opened 2023-11-15
Add support for Bedrock Cohere embeddings


danielmitchell:
Bedrock has recently added support for Cohere embeddings. It would be great if this app could support the English and multilingual Cohere embeddings models.

https://aws.amazon.com/blogs/aws/amazon-bedrock-now-provides-access-to-cohere-command-light-and-cohere-embed-english-and-multilingual-models/

Gmreh:
Yes, that would be indeed an extremely beneficial update...

PR #361, opened 2024-02-05
Add support for cohere/titan embedding models

QuinnGT:
*Issue #, if available:*

Issues open that this resolves, #218 

*Description of changes:*

Brought over changes from spugachev 3 month old branch on cohere. Added changes for new schema. Deployed and tested in fresh environment last night but I see massi-ang opened #359 this morning. However cohere embedding needs additional changes for how it handles inputs. Documented [here](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html).

search_document type and search_query type are used. Classification and clustering are not needed at this time. 

[embedding_types](https://docs.cohere.com/reference/embed) was not defined as float is appropriate for this implementation. 


lib/shared/layers/python-sdk/python/genai_core/embeddings.py
```diff
@@ -73,30 +73,47 @@ def _generate_embeddings_openai(
...
     bedrock = genai_core.clients.get_bedrock_client()
 
     if not bedrock:
         raise genai_core.types.CommonError("Bedrock is not enabled.")
 
     ret_value = []
-    for value in input:
-        body = json.dumps({"inputText": value})
+    model_provider = model.name.split(".")[0]
+    if model_provider == "amazon":
```
bigadsoleiman, line 85: 2 suggestions:
- Let's have providers in [types.py](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/shared/layers/python-sdk/python/genai_core/types.py) instead of hardcoded at each condition. 
- better to break down the provider-related code into separate:
  - `_generate_embeddings_amazon`
  - `_generate_embeddings_cohere` or any better naming idea 
  and just have the condition statement in this main `_generate_embeddings_bedrock`

lib/shared/layers/python-sdk/python/genai_core/embeddings.py
```diff
@@ -73,30 +73,47 @@ def _generate_embeddings_openai(
...
+            embedding = response_body.get("embedding")
+
+            ret_value.append(embedding)
+
+        ret_value = np.array(ret_value)
+        ret_value = ret_value / \
+            np.linalg.norm(ret_value, axis=1, keepdims=True)
+        ret_value = ret_value.tolist()
+    elif model_provider == "cohere":
+        input_type = "search_query" if task == "retrieve" else "search_document"
```
bigadsoleiman, line 104: same for tasks let's type instead if hardcode them

bigadsoleiman (changes requested):
Thanks for this submission!

Just added a couple of feedbacks

bin/config.ts
```diff
@@ -52,6 +52,16 @@ export function getConfig(): SystemConfig {
           provider: "bedrock",
           name: "amazon.titan-embed-text-v1",
           dimensions: 1536,
+        },
+        {
```
bigadsoleiman, line 56: please include [titan](https://github.com/aws-samples/aws-genai-llm-chatbot/pull/359/files#diff-45deac45daa48ef2d176c1d42cb9629f2607b020c8ea4bd8ab6763ac6033ac5fR37) from https://github.com/aws-samples/aws-genai-llm-chatbot/pull/359 which i closed in favor of this PR

bigadsoleiman (changes requested):

bin/config.ts
QuinnGT: Added

lib/chatbot-api/functions/api-handler/routes/embeddings.py
massi-ang, line 1: let's use an Enum for task

massi-ang (changes requested):
Thanks for the PR.

bin/config.ts
QuinnGT: Actually one thought I had on this @bigadsoleiman is we are using Unstructured to to handle ingestion and partition all content, or at least I think it is. Would this conflict with Unstructured partition_image with ocr? I'll look into it and see if there is overlap on functions.

lib/shared/layers/python-sdk/python/genai_core/opensearch/query.py
```diff
@@ -44,7 +44,7 @@ def query_workspace_open_search(
         raise genai_core.types.CommonError("Cross encoder model not found")
 
     query_embeddings = genai_core.embeddings.generate_embeddings(
-        selected_model, [query]
+        selected_model, [query], "retrieve"
```
bigadsoleiman, line 47: use `Task.RETRIEVE.value`

lib/shared/layers/python-sdk/python/genai_core/aurora/query.py
```diff
@@ -45,7 +45,7 @@ def query_workspace_aurora(
         raise genai_core.types.CommonError("Cross encoder model not found")
 
     query_embeddings = genai_core.embeddings.generate_embeddings(
-        selected_model, [query]
+        selected_model, [query], "retrieve"
```
bigadsoleiman, line 48: use `Task` enum

lib/shared/layers/python-sdk/python/genai_core/chunks.py
```diff
@@ -40,7 +40,7 @@ def add_chunks(
         raise genai_core.types.CommonError("Embeddings model not found")
 
     chunk_embeddings = genai_core.embeddings.generate_embeddings(
-        embeddings_model, chunks
+        embeddings_model, chunks, "store"
```
bigadsoleiman, line 43: use `Task` enum

lib/shared/layers/python-sdk/python/genai_core/workspaces.py
```diff
@@ -112,7 +112,7 @@ def create_workspace_aurora(
     if not embeddings_model:
         raise genai_core.types.CommonError("Invalid embeddings model")
     # Verify that the embeddings model
-    genai_core.embeddings.generate_embeddings(embeddings_model, ["test"])
+    genai_core.embeddings.generate_embeddings(embeddings_model, ["test"], "store")
```
bigadsoleiman, line 115: use `Task` enum

lib/shared/layers/python-sdk/python/genai_core/workspaces.py
```diff
@@ -179,7 +179,7 @@ def create_workspace_open_search(
...
```
bigadsoleiman, line 182: use `Task` enum

bigadsoleiman (changes requested):

QuinnGT:
Quick update. I pushed updates for everything except for titan multimodal. Currently deploying those changes and testing before taking on titan multimodal. 

In regards to [Titan Multimodal Embeddings](https://docs.aws.amazon.com/bedrock/latest/userguide/titan-multiemb-models.html), it will need a bit more than just calling out the additional model. It will need support for it's API payload like max input text tokens and max input image size. However we are not capturing images in our use of Unstructured or rather lack of use of unstructured, like the `[partition_image](https://unstructured-io.github.io/unstructured/core/partition.html#partition-image)` function. Also I see unstructured is poorly used as we pull the full container yet only use the loader. 

As soon as the cohere changes finish and tests pass, I'll circle back on titan multimodal and unstructured. Feel free to post thoughts  or feedback.

massi-ang:
`maxInputTokens` is optional and we currently only need to support text embeddings. This was already tested in #359 and working fine.

QuinnGT:
Alright, all requested changes have been made. I also fixed the embedding prompt to show when enableRag was true vs just being skipped no matter what in the magic-config.

New environment has been deployed. Bedrock and Sagemaker embedding models tested as functional.

QuinnGT:
Found issue with creating workspace. Not sure if it's from me or changes merged into this PR. Need to test and see.

cli/magic-config.ts
```diff
@@ -322,15 +338,22 @@ async function processCreateOptions(options: any): Promise<void> {
...
+      validate(value: string) {
+        if ((this as any).state.answers.enableRag) {
+          return value ? true : 'Select a default embedding model'; 
+        }
+      
+        return true;
       },
-    },
+      skip() {
+        return !answers.enableRag;
```
Rob-Powell, line 354: this is wrong as if you select only kendra as your RAG you dont need an embedding model.

QuinnGT: All good, fair point. Made a change so it only shows when it's aurora and/or opensearch.

QuinnGT:
Resolved the latest issues. Feel free to test and let me know if anything else is needed.

lib/shared/layers/python-sdk/python/genai_core/embeddings.py
```diff
@@ -97,12 +107,25 @@ def _generate_embeddings_bedrock(
     ret_value = np.array(ret_value)
     ret_value = ret_value / np.linalg.norm(ret_value, axis=1, keepdims=True)
     ret_value = ret_value.tolist()
-
     return ret_value
 
 
+def _generate_embeddings_cohere(model: EmbeddingsModel, input: List[str], task: Task, bedrock):
+    input_type = "search_query" if task == Task.RETRIEVE else "search_document"
```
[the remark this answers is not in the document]
QuinnGT: Sorry I missed that, good catch. I'll update it today.

QuinnGT:
@bigadsoleiman I noticed an error in deployment that's happening for log groups due to size of name. I tested it  [here ](https://github.com/QuinnGT/aws-genai-llm-chatbot/commit/ba191410976277f13ea69aaddef463962d086964)and it works. Do you want me to bring that into this PR or wait till this is approved?

massi-ang:
> @bigadsoleiman I noticed an error in deployment that's happening for log groups due to size of name. I tested it [here ](https://github.com/QuinnGT/aws-genai-llm-chatbot/commit/ba191410976277f13ea69aaddef463962d086964)and it works. Do you want me to bring that into this PR or wait till this is approved?

Hi Quinn, I checked your change and my concern is that if you deploy two stacks, they will share the same log group, since the value of `id` will be the same in for both.

QuinnGT:
> > @bigadsoleiman I noticed an error in deployment that's happening for log groups due to size of name. I tested it [here ](https://github.com/QuinnGT/aws-genai-llm-chatbot/commit/ba191410976277f13ea69aaddef463962d086964)and it works. Do you want me to bring that into this PR or wait till this is approved?
> 
> Hi Quinn, I checked your change and my concern is that if you deploy two stacks, they will share the same log group, since the value of `id` will be the same in for both.

Great point @massi-ang . How about I use the config.prefix with the id so it looks something like `logGroupName: /aws/vendedlogs/states/${config.prefix}-${id}`. 

Before: /aws/vendedlogs/states/CreateAuroraWorkspace-CreateAuroraWorkspaceSMLogGroup
After: /aws/vendedlogs/states/dev-CreateAuroraWorkspaceSMLogGroup

massi-ang:
I would suggest using `/aws/vendedlogs/states/${cdk.Stack.of(this).name}-${id}`, but there could still be issues in case we apply the same solution to other constructs which have the same `id` (since `id` in only unique in the scope of the parent construct).

QuinnGT:
@bigadsoleiman is there anything else pending that I missed? If not, can you approve this PR? Thank you!

bigadsoleiman (approved):

Cerrix:
Hi @QuinnGT,
I tested the pull request because I was looking for the Cohere Multilingual implementation. When I selected that model for embeddings, I received the following error response from AppSync:
`{
  "data": null,
  "errors": [
    {
      "path": [
        "calculateEmbeddings"
      ],
      "data": null,
      "errorType": "TypeError",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "Object of type Task is not JSON serializable"
    }
  ]
}`
both creating the workspace or testing the embedding. 
With Titan everything is fine.

Thank you!

QuinnGT:
> Hi @QuinnGT, I tested the pull request because I was looking for the Cohere Multilingual implementation. When I selected that model for embeddings, I received the following error response from AppSync: `{ "data": null, "errors": [ { "path": [ "calculateEmbeddings" ], "data": null, "errorType": "TypeError", "errorInfo": null, "locations": [ { "line": 2, "column": 3, "sourceName": null } ], "message": "Object of type Task is not JSON serializable" } ] }` both creating the workspace or testing the embedding. With Titan everything is fine.
> 
> Thank you!

@Cerrix  thank you for reporting this. Can you share what RAG you were testing with?

Cerrix:
Yep: I’m testing in with Open Search (and just that). I get the same error even if testing cohere through the “embeddings” page (not only creating a new workspace).

thank you

QuinnGT:
> Yep: I’m testing in with Open Search (and just that). I get the same error even if testing cohere through the “embeddings” page (not only creating a new workspace).
> 
> thank you

Hey @Cerrix I enabled and deployed an opensearch instance and I'm not seeing any errors with workspace creation or testing embeddings. I'm not sure if it's a new or existing deployment you are testing in but try doing a fresh deployment from the latest, along with a new npm install and build. Feel free to report back with more information to troubleshoot, if needed.

QuinnGT:
Comparison of embedding models from demo data. 
image

Cerrix:
It was a week old deploy. But let me try again from scratch! I'll keep you posted!

Cerrix:
It is so strange. I re-deployed from scratch the repository branching your PR in Oregon (us-west-2).

Here is my config file: 
```
{
  "prefix": "llmchatbot-demo",
  "privateWebsite": false,
  "certificate": "",
  "domain": "",
  "bedrock": {
    "enabled": true,
    "region": "us-west-2"
  },
  "llms": {
    "sagemaker": []
  },
  "rag": {
    "enabled": true,
    "engines": {
      "aurora": {
        "enabled": false
      },
      "opensearch": {
        "enabled": true
      },
      "kendra": {
        "enabled": false,
        "createIndex": false,
        "external": [],
        "enterprise": false
      }
    },
    "embeddingsModels": [
      {
        "provider": "sagemaker",
        "name": "intfloat/multilingual-e5-large",
        "dimensions": 1024
      },
      {
        "provider": "sagemaker",
        "name": "sentence-transformers/all-MiniLM-L6-v2",
        "dimensions": 384
      },
      {
        "provider": "bedrock",
        "name": "amazon.titan-embed-text-v1",
        "dimensions": 1536
      },
      {
        "provider": "bedrock",
        "name": "amazon.titan-embed-image-v1",
        "dimensions": 1024
      },
      {
        "provider": "bedrock",
        "name": "cohere.embed-english-v3",
        "dimensions": 1024
      },
      {
        "provider": "bedrock",
        "name": "cohere.embed-multilingual-v3",
        "dimensions": 1024,
        "default": true
      },
      {
        "provider": "openai",
        "name": "text-embedding-ada-002",
        "dimensions": 1536
      }
    ],
    "crossEncoderModels": [
      {
        "provider": "sagemaker",
        "name": "cross-encoder/ms-marco-MiniLM-L-12-v2",
        "default": true
      }
    ]
  }
}
```
Still I got the same error:
Screenshot 2024-03-04 at 00 06 01

QuinnGT:
@Cerrix  Interesting, I'm not sure. It looks like it's an error from the appsync function. Almost like the old version of app sync functions are trying to be used with the newer genai_core python functions. Did you purge your cdk.out when doing a fresh deployment? Also double check cdk version is updated on your local as there seems to be some rebuild bug introduced on each release.

@massi-ang I see Bigad approved the PR but we're still waiting on you or another repo owner for approval. What else do you need to from me?

Cerrix:
I'm deploying following the Cloud9 instructions of the sample documentation: https://aws-samples.github.io/aws-genai-llm-chatbot/guide/deploy.html

If we need to update some dependencies I think it must be documented before merging the PR in main.

I can try with a vanilla cloud9 instance but we need to double check the issue. I don't know if someone from the team would like to try the deploy!

lib/shared/layers/python-sdk/python/genai_core/types.py
```diff
@@ -31,6 +31,8 @@ class Provider(Enum):
     BEDROCK = "bedrock"
     OPENAI = "openai"
     SAGEMAKER = "sagemaker"
+    AMAZON = "amazon"
+    COHERE = "cohere"
```
massi-ang, line 35: This is confusing semantic as Bedrock and Sagemaker are not Provider in the same sense as Amazon and Cohere are. Please split Model Provider from Hosting Providers.

massi-ang (changes requested):
Please fix the `Provider` type with the right semantics

lib/shared/layers/python-sdk/python/genai_core/types.py
QuinnGT: I agree, it's not clean. I'll split it and push today.

QuinnGT: Splitting the providers into `ModelProviders` and `HostingProviders` has a way bigger impact than I expected. `Provider `is widely used and `ModelProvider` is oddly used in the react-app/common/types. At this point I've made changes in 20 files but the sprawl of changes needed continues. 

Do you want me to continue to do the split or am I just approaching this incorrectly and overcomplicating it?

bigadsoleiman: @QuinnGT while I agree there should be a separation the give the ramification of this split I'd say is out of the scope of this PR. So if you want to revert back to just provider we can create a separate issue for this refactor.

QuinnGT: @bigadsoleiman okay, I didn't commit my changes so it should all be stable as is now. Fork is sync'd with main now. 

You can approve the PR and close it.

bigadsoleiman (approved):

Issue #227, opened 2023-11-20
Make workspace private


abduljaleel:
Currently workspace is not private for a user account.  Its public and visible across all users.   Workspace should be private and option to make it public.

hghandri:
That feature could be very useful for many organization. +1

flamingquaks:
We are aiming to start this effort following some backend refactoring and work on #99 being implemented. 

I'm evaluating use cases around this effort now and will drop a proposal as a comment for folks soon.

hghandri:
Hi,

I made some changes into fork project to add private workspace capability.
I plan to push PR with all changes and let you see what could you do with.

Best regards,

abduljaleel:
> Hi,
> 
> 
> 
> I made some changes into fork project to add private workspace capability.
> 
> I plan to push PR with all changes and let you see what could you do with.
> 
> 
> 
> Best regards,

Looking forward to this, happy to be a tester for this feature, private workspace is key.

flamingquaks:
> Hi,
> 
> I made some changes into fork project to add private workspace capability. I plan to push PR with all changes and let you see what could you do with.
> 
> Best regards,

+1 Looking forward to what you've come up with!

hghandri:
Hi,

here my modest contribution : https://github.com/aws-samples/aws-genai-llm-chatbot/pull/284

I hope that could be useful for us.

++

hghandri:
Hello,

I've just pushed a new version compliant with last project changes.
Please let me know what do you think.

PR #376, opened 2024-02-20
Let workspace private and personal by default for all user

hghandri:
Issue: https://github.com/aws-samples/aws-genai-llm-chatbot/issues/227

Implement private workspace, every user create a workspace now is only private and not shareable with other user.
This feature implement a new DynamoBD table handle policies on top of Workspace Dynamodb table.
This implementation could facilitate sharing workspaces between users with writeable / readable policy  option through sharing process like Google Drive


massi-ang:
Thanks for submitting this PR. Is this changing all Workspaces to private to the user?

hghandri:
Yes by default, all workspace will be private.
If we go ahead of this feature we could add interface to set if we want to open publicly the workspace or for some people ( example fetch a user from cognito )

massi-ang:
As this is a breaking change, we should either think about an opt-in/out settings or the possibility to set workspaces as public/private and retain all existing workspaces as public.

hghandri:
I will do a migration script to let all existing workspace public.

On further in case the workspace does not have owner ( legacy workspace ) the first one who will set through interface that the workspace is private then he will became the owner of this workspace for ever.

What do you think ?

hghandri:
I closed this PR and opened a new one with more impact.

PR #384, opened 2024-02-23
feat: adding schedule option for sagemaker models

Rob-Powell:
*Issue #, if available:*

*Description of changes:*

This change implements the ability to set a schedule via event bridge to toggle on and off the sagemaker hosted models. Typically the solution is consumed during business hours so enabling a schedule to suit those hours makes sense. See updated docs for more information.

I had to build a class to validate the peculiar cron expression format that Event Bridge supports to help give feedback to users inputting a cron expression early, rather than waiting for a deployment to fail later.

### DIFF
--- a/cli/aws-cron-expressions.ts
+++ b/cli/aws-cron-expressions.ts
@@ -1,0 +1,60 @@
+export const minuteExp = `(0?[0-9]|[1-5][0-9])`; // [0]0-59
+export const hourExp = `(0?[0-9]|1[0-9]|2[0-3])`; // [0]0-23
+export const dayOfMonthExp = `(0?[1-9]|[1-2][0-9]|3[0-1])`; // [0]1-31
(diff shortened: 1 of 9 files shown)


massi-ang:
Thank for contributing this much needed feature that would enable to reduce the costs of the solution.
Overall I do not see any issues with the implementation but I would like to see a more guided way to provide schedules.
Can you modify the config script to allow the user to:
- specify a start and end date for the schedule - outside of this period the endpoint should be turned off
- select the the days to which the schedule applies to (multiselect)
- specify the start and end time in HH:MM format 
- derive the cron expressions from the user input

Rob-Powell:
@massi-ang have completed the changes, I also added a TZ feature to make it even easier for users to set a schedule.

massi-ang (approved):
I am happy with the changes.

PR #396, opened 2024-03-06
Multimodal chat for Claude

### LINKED ISSUE #395: Enable multimodal chat with Claude 3 models
Claude 3 models supports image modality and they can be used for the multimodal chat in addition to Idefics

massi-ang:
Redesign of the Idefics Model Interface to accept multiple model via adapters. Added Claude 3 in addition to Idefics

*Issue #, if available:*
Resolves #395

*Description of changes:*
- refactoring of the Idefics model interface to support multiple model via adapters
- refactored Idefics code into an adapter
- implemented adapter for Claude 3 
- moved the `registry` functionality into `genai_core`.


lib/model-interfaces/idefics/functions/request-handler/adapters/claude.py
```diff
@@ -0,0 +1,104 @@
...
+            "content": [{"type": "text", "text": prompt}],
+        }
+        prompts.append(user_msg)
+        for file in files:
+            user_msg["content"].append(get_image_message(file))
+
+        return json.dumps(
+            {
+                "anthropic_version": "bedrock-2023-05-31",
+                "max_tokens": 512,
```
bigadsoleiman, line 65: are hardcoded `max_tokens` and `temperature` really needed here?

lib/aws-genai-llm-chatbot-stack.ts
```diff
@@ -100,46 +100,45 @@ export class AwsGenAILLMChatbotStack extends cdk.Stack {
...
-            ),
-            modelInterface: sns.FilterOrPolicy.filter(
-              sns.SubscriptionFilter.stringFilter({
-                allowlist: [ModelInterface.Idefics],
-              })
-            ),
-          },
-        })
-      );
+    const ideficsInterface = new IdeficsInterface(this, "IdeficsInterface", {
```
bigadsoleiman, line 111: Wouldn't be better to also change the construct name to `MultimodalInterface` ?

bigadsoleiman (commented):
Should we change also the construct name to `x

lib/model-interfaces/idefics/functions/request-handler/adapters/claude.py
massi-ang: Just replicated this: https://github.com/aws-samples/aws-genai-llm-chatbot/blob/0545d45e4a1f7f321a3247f5988da96b2b37a805/lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/base.py#L36-L37

massi-ang: These values will be overridden in `handle_run`.

lib/aws-genai-llm-chatbot-stack.ts
massi-ang: I did that first, but updating an existing stack will take ages, so I decided to keep the old name.

bigadsoleiman (approved):

Issue #390, opened 2024-03-04
Mixtral not available in bedrock adapter


jeremylatorre:
Would be great if we could test the new Mixtral 8x7B newly available

PR #400, opened 2024-03-07
Add support for Mistral models via Bedrock

massi-ang:
*Issue #, if available:*

Resolves: #390 

*Description of changes:*
Given that the prompt format of Mistral and Mixtral are identical to Llama2, we reused the Llama2 adapter registering it also for Mistral models

The input and output adaptations are different and taken care by the base Bedrock LLM class

Finally, we revised the LLama2 prompts.


lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/base.py
```diff
@@ -33,7 +29,9 @@ def prepare_input(
         if provider == "anthropic":
```
bigadsoleiman, line 29: should we move model providers to enum?

bigadsoleiman (approved):

Issue #298, opened 2024-01-10
Possibility to use custom domain name  with Cloudfront


hghandri:
Hi,

do you have plan about managing a custom domain name with Cloudfront with secure protocol HTTPS of course.

Thanks.

hghandri:
Hi 
I could implement this feature, it's  possible to talk about it at anytime to go further.

++

azaylamba:
@hghandri A workaround for this is to add alternate domain name in cloudfront distribution.

PR #416, opened 2024-03-14
feat: Possibility to use custom domain name with Cloudfront

JeromeGuyon:
*Issue #298: Possibility to use custom domain name with Cloudfront*

*Description of changes:*
- add possibility to configure a custom domain for the public cloudfront
- add corresponding documentation


cli/magic-config.ts
```diff
@@ -170,22 +170,42 @@ async function processCreateOptions(options: any): Promise<void> {
...
       skip(): boolean {
-        return !(this as any).state.answers.privateWebsite;
+        return !(this as any).state.answers.privateWebsite && !(this as any).state.answers.customPublicDomain;
       },
     },
     {
       type: "input",
       name: "domain",
-      message: "Domain for private website",
+      message(): string {
```
bigadsoleiman, line 200: What about asking for Route53 Zone Id too so we can create the [A record with CDK directly](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_route53-readme.html)?

JeromeGuyon: That's a good point.
Most of the time Route53 public zone is centralized in an AWS Account. 
This does not limits this possibility (local or central).

I propose to create another PR for this, 
[Domain Names and Certificates](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#domain-names-and-certificates)
with a migration to [CloudFrontWebDistribution API](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cloudfront-readme.html#cloudfrontwebdistribution-api)

bigadsoleiman: Agreed

bigadsoleiman (approved):

Issue #450, opened 2024-04-10
S3 Bucket Name is Empty when deploying UI Bucket


gknorman-amazon:
I tried deploying today with minimal input (with a private website only accessible in a VPC) during the config process, but the deployed CFN had an empty string for the S3 Bucket Name for the UI hosting bucket:

"UserInterfaceWebsiteBucket2BDEA247": {
   "Type": "AWS::S3::Bucket",
   "Properties": {
    **"BucketName": "",**
    "LoggingConfiguration": {
     "DestinationBucketName": {
      "Ref": "UserInterfaceWebsiteLogsBucket54728780"
     }
    }
    
 I didn't provide a certificate ARN nor a domain name, and I'm not sure if these are necessary to generate the Bucket Name, but I expected there to be an explicit step to name the bucket during configuration if input is actually needed from the user.

gknorman-amazon:
In the meantime, I'll fix this the "wrong" way by creating a default bucket name in case the domain name is left empty during the config process in `user-interface/index.ts`

`const defaultBucketNameString = "website-bucket" + cdk.Stack.of(this).account + "-" + cdk.Stack.of(this).region; `
    
 `const defaultDomainName = props.config.domain ? props.config.domain : defaultBucketNameString;`

If I enter my own domain name but I haven't set up my own hosted zone, etc already, will it create it for me?  If so, I feel very silly and would ask for an update in config process or the documentation to make us aware of this.

Rob-Powell:
As per the docs you need to do the pre req steps and then pass those values to the installer:

https://aws-samples.github.io/aws-genai-llm-chatbot/documentation/private-chatbot.html

gknorman-amazon:
Can these prereqs be shown / linked to on the Deployment Guide page then?

This is the first time seeing this "Private Chatbot" page since I went straight from the "The Project" page and the "Deployment Guide" link at the bottom of that page.  I didn't realize there were any prereqs when I started the Cloud 9 deployment steps.

Rob-Powell:
They are linked to on the left hand side navigation. but there should be validation on those fields in config creation.

PR #451, opened 2024-04-11
fix: logic for private website and add validation on config params

Rob-Powell:
*Issue #, if available:*

#450 

*Description of changes:*

ACM cert and domain name entries are required when selecting private website or when creating a public website with a custom domain. this change adds validation to those fields to make them required in those circumstances and does some additional validation to ensure a valid ACM ARN is supplied.


ystoneman (commented):
Looks great to me! This doesn't add additional functionality -- just adds validation that should be there.

bigadsoleiman (approved):

Issue #387, opened 2024-02-26
Connect the solution with SAML SSO


ahmedseeif2:
We're aiming to enable users to sign in using their Azure AD credentials as a SAML provider. I've been following the documentation on how to connect SAML SSO to AWS Cognito, and after conducting some testing, it seems that everything is functioning as expected. However, I've noticed that the UI doesn't currently provide an option to use Azure AD as an authentication method. Can we update the code to add this option? When I try to login with the user that authenticated by SAML it errors out with: User doesn't exist

PR #460, opened 2024-04-20
feat: Support for cognito federated indentities

Rob-Powell:
*Issue #, if available:*

#387 

*Description of changes:*

Users of the solution usually want to quickly demo and highlight value, connecting to an organisations existing identity provider is a quick way of enabling secure access to the solution to a defined user base.

Adds support for users to configure and leverage the solutions ability to Federate user logon via supported Cognito federation.

I have written docs for this solution see additions under docs/

Currently this solution only supports: 

- Cognito custom OIDC  providers.
- Cognito custom SAML providers.
- Supports setting the configuration `later` within the setup process but a helpful guided process is also provided.

2 Custom Resource lambda functions had to be used"

- 1 to update the cognito client with the created domain (circular dependency)
- 2 to securely pass the OIDC client secret to the cognito identity provider avoiding having the secret stored in cloudformation


bigadsoleiman (approved):

bigadsoleiman:
This is great, much needed addition @Rob-Powell

Issue #149, opened 2023-10-26
Feature request: ability to delete/erase domain data added to RAG workspace


saragerion:
### Summary

It could be useful to be able to selectively delete specific domain data added in the RAG workspace.
Domain data I am referring to includes:
- Files uploaded 
- Text
- Q/A
- Websites

### Motivation
There are some use cases in which a company may want or need to delete specific domain data after it has been added to the RAG storage.  This can happen for example when companies need to adhere to compliance requirements and therefore should be able to erase private data from their workloads when requested.
Example: [European Union’s General Data Protection Regulation (GDPR)](https://aws.amazon.com/compliance/gdpr-center/) “right to erasure” or “right to be forgotten” requirements, when applicable.

ajaylamba-provar:
Or the more common use case would be - the content has been updated.

azaylamba:
@flamingquaks @massi-ang I'm trying to make changes for this, could you guys give some pointers to start with please?

azaylamba:
I am working one this one. I have successfully implemented this for OpenSearch but my fork is no in sync with main after migration to AppSync. 
I will sync with main and will modify the code accordingly.

azaylamba:
Update: Changes are done, testing in progress.

azaylamba:
Finally, raised PR for this.

azaylamba:
Changes for this have been merged to main, this issue can be closed as done.

PR #464, opened 2024-04-20
feat: Added delete document functionality

azaylamba:
*Issue #149 :*

*Description of changes:* Added delete functionality for all types of documents (Files, Texts, Q&A and Websites). The feature deletes the documents from S3 upload bucket, S3 processed bucket, DynamoDB documents table, OpenSearch index and also updates DynamoDB workspaces table. Following are the major code changes:
1. Added delete button on UI for each row of the documents.
2. Added confirmation dialog via Modal so that user can Cancel/Delete the document from there.
3. Created AWS step function to use State Machines and delete document workflow. This way, the whole process is organised and is automatically rolled back if any of the operation in the step function fails.

Major components and their working is as below:
1. documents-tab.tsx has functionality related to delete button and handling of confirmation Modal.
2. documents-client.ts has function deleteDocument to hit the backend API.
3. delete_document function in lib/chatbot-api/functions/api-handler/routes/documents.py handles the API request
4. deleteDocumentWorkflow is created in lib/rag-engines/workspaces/index.ts
5. delete-document.ts has internal structure of Delete document workflow
6. The lambda function to handle the workflow is written in lib/rag-engines/workspaces/functions/delete-document-workflow/delete/index.py
7. The execution of state machine starts in delete_document function of lib/shared/layers/python-sdk/python/genai_core/documents.py
8. The actual deletion of documents happens in delete_open_search_document function of lib/shared/layers/python-sdk/python/genai_core/opensearch/delete.py

Request flow would be like documents-client -> documents.py (api handler) -> documents.py (genai_core) -> index.py (delete-document-workflow) -> delete.py (genai_core/opensearch)

As part of this change, also updated version of opensearch-py which was initially updated as calling direct http methods was not allowed in earlier version but later on calling http methods was not required. Kept this change for future perspective as it would have no impact.


azaylamba:
@massi-ang @bigadsoleiman Could you please review this?

lib/shared/layers/python-sdk/python/genai_core/aurora/delete.py
```diff
@@ -68,3 +73,65 @@ def delete_aurora_workspace(workspace: dict):
...
+        print(f"An error occurred: {error}")
+
+def deleteAuroraDocument(document_id: str, table_name: str):
+    try:
+        with AuroraConnection(autocommit=False) as cursor:
+            cursor.execute(
+                sql.SQL("DELETE FROM {table} WHERE document_id = %s").format(table=table_name),
+                (document_id,),
+            )
+            print(f"Deleted document {document_id} from {table_name}")
```
gbone-restore, line 135: There is a missing cursor commit. Add cursor.connection.commit()

azaylamba:
@bigadsoleiman @massi-ang would you be able to have a look at this PR?

bigadsoleiman:
Thanks @azaylamba for this. Looking into it

bigadsoleiman:
Good one and well detailed!

Tested with all engines and against an existing stack to ensure retro-compatibility with already uploaded docs.
Happy to merge.

PR #487, opened 2024-05-13
Fix resource naming

massi-ang:
*Issue #, if available:*
Some resources had inconsistent naming making harder to understand resource dependencies

*Description of changes:*
- change resource name
- removed unnecessary logging found while investigating the other issue


massi-ang:
This resource name change does not affect stateful resources (the Kendra index), ie it is safe to be applied to existing deployments.

bigadsoleiman (approved):

PR #509, opened 2024-07-02
Fix Kendra character count validation error

anaseol:
*Issue #, if available:*

Kendra retrieval fails when the query has more than 1000 characters.
The error can be reproduced by choosing a workspace that uses the Kendra engine and sending a message longer than 1000 characters.

kendra_character_count_validation_error

*Description of changes:*
An attempt to solve this problem would be to remove any leading and trailing whitespace characters and send only the first 1000 characters as query text.


Rob-Powell:
I'm not sure this is the best approach as it can also be solved by increasing the service limit for kendra, perhaps this should be some docs guidance to increase the service quota instead?

PR #518, opened 2024-07-24
fix(api): Fixed getDocument(s) endpoint errors

flamingquaks:
When attempting to view details of an RSS feed, the getDocument endpoint was throwing an error around crawler_properties, potentially due to a merge in the past. The updated code properly handles if crawler properties are applied. 


Rob-Powell:
LGTM - thanks for the fix

bigadsoleiman (approved):

PR #520, opened 2024-07-25
Feature/hide not enabled models 

eldimi:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/models/index.ts
+++ b/lib/models/index.ts
@@ -34,6 +34,7 @@ export class Models extends Construct {
     super(scope, id);
 
     const models: SageMakerModelEndpoint[] = [];
+    const bedrockEnabledModels: string[] = ['anthropic.claude-3-haiku-20240307-v1:0', 'anthropic.claude-3-sonnet-20240229-v1:0'];
 
     let hfTokenSecret: secretsmanager.Secret | undefined;
     if (props.config.llms.huggingfaceApiSecretArn) {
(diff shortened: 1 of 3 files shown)


lib/models/index.ts
```diff
@@ -29,11 +29,13 @@ export interface ModelsProps {
 export class Models extends Construct {
   public readonly models: SageMakerModelEndpoint[];
   public readonly modelsParameter: ssm.StringParameter;
+  public readonly bedrockEnabledModelsParameter: ssm.StringParameter;
 
   constructor(scope: Construct, id: string, props: ModelsProps) {
     super(scope, id);
 
     const models: SageMakerModelEndpoint[] = [];
+    const bedrockEnabledModels: string[] = ['anthropic.claude-3-haiku-20240307-v1:0', 'anthropic.claude-3-sonnet-20240229-v1:0'];
```
massi-ang, line 38: This value should come from the `config.json` file. You need to add a new property to the `SystemConfig` interface to store these values and add a multiselect option to the magic config to let the user choose which models to enable.

massi-ang (changes requested):
Nice feature. Please add the possibility to choose the models to be enabled via the config tool instead of hard coded values in the CDK construct.

lib/models/index.ts
eldimi: Should we add a free text with comma seperated model ids or a multi-selected to select from all the available foundation models ? I would vote for free text, since it doesn't need maintance - adding new models - and a bit more user-friendly - user will type instead of scrolling through the models to select the one's needed.

massi-ang: I you just want to allow-list few models, a text field could work fine, but you would still need to validate the modelIds. I think a multiselect is more user friendly and you can always populate the values dynamically by querying the bedrock API. Anyway, to add new models we also need to add them to the rest of the code.

Issue #501, opened 2024-06-11
Erratic streaming of text


aD-aws:
The text displayed while streaming is quite erratic. Attached is an example.
[AWS GenAI Chatbot.webm](https://github.com/aws-samples/aws-genai-llm-chatbot/assets/172373760/33c08b1b-e30d-4fa4-b4b1-c7213fafb9c9)

charles-marion:
Fixed in #528

PR #528, opened 2024-08-02
fix: Token order during streaming

charles-marion:
*Issue #, if available:*
#501

*Description of changes:*
Fixes two issues:
* Only shows a token if the previous ones are available (so it only adds the words at the end)
* Fix token missed. This happens when receiving 2 tokens in parallel. The fix is to move the reference of `lastMessage.tokens` out of `updateMessageHistoryRef` (so it uses the same array)

https://github.com/user-attachments/assets/4e4b5a6b-a67d-40a0-9d88-8181244b541b


bigadsoleiman (approved):

bigadsoleiman:
LGTM.

Thanks for this long awaited fix

PR #530, opened 2024-08-05
Update the path to the magic config executable

massi-ang:
*Issue #, if available:*

*Description of changes:*
Updated the path to the magic config executable after a change in `tsconfig.json` introduced in  #527 


charles-marion (approved):

charles-marion:
Thank you for the fix!

Issue #543, opened 2024-08-12
Models dropdown shows `azure:openai`even if no models have been configured


massi-ang:
I have pulled the latest commit and re-deployed the chatbot on an existing installation. 
Now I am getting

<img width="463" alt="image" src="https://github.com/user-attachments/assets/4c292a4a-4ecb-44db-a091-5bfa0525a68c">

at the top of my models list.

grinko:
Fixed 
<img width="533" alt="Screenshot 2024-08-14 at 11 13 49 AM" src="https://github.com/user-attachments/assets/464daf10-807e-4576-b1c5-b885f382e6c7">

PR #546, opened 2024-08-14
fix: Models dropdown shows azure:openai even if no models have been c…

grinko:
fix: Models dropdown shows azure:openai even if no models have been configured 

*Issue #543

*Description of changes:*
* fix: Models dropdown shows azure:openai even if no models have been configured 
* Update deployment instructions


lib/shared/layers/python-sdk/python/genai_core/models.py
```diff
@@ -57,7 +57,8 @@ def list_azure_openai_models():
     # azure openai model are listed, comma separated in
     # AZURE_OPENAI_MODELS variable in external API secret
     models = genai_core.parameters.get_external_api_key("AZURE_OPENAI_MODELS") or ""
-
+    if not models:
+        return []
```
charles-marion, line 61: I would None to be consistent with the other methods in this file.

docs/guide/deploy.md
```diff
@@ -112,6 +112,7 @@ npm run test && pip install -r pytest_requirements.txt && pytest tests
 ```bash
 npm run config
 ```
+Note, that `anthropic.claude-instant-v1` (Claude Instant) and `amazon.titan-embed-text-v1` (Titan Embeddings G1 - Text) need to be enabled in Bedrock before you run the command.
```
charles-marion, line 115: I would not specify models here. There are the model needed for integ tests.

On line 119 below  I would update instead

`we support all models provided by Bedrock` to `we support all models provided by Bedrock that [were enabled](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html)`

lib/shared/layers/python-sdk/python/genai_core/models.py
grinko: Updated

docs/guide/deploy.md
grinko: Updated

charles-marion (approved):

PR #569, opened 2024-09-17
feat: Enable usage token and use Converse API with all Bedrock models

charles-marion:
*Issue #, if available:*
Follow up of https://github.com/aws-samples/aws-genai-llm-chatbot/pull/564

*Description of changes:*
Since the bedrock [converse api](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html) allows a team to "write code once and use it with different models.", this change aims to use the same code for all Bedrock models.

The only differences are some models do not support system prompt or streaming.

The main goal of this change is to add usage token tracking to all Bedrock models.

**Changes**
- Remove model specific code for Bedrock
- Add unit test for the bedrock adapters
- Add an user friendly message is a bedrock model is not enabled
- Add unit test

**Notes**
- the main changes are in lib/model-interfaces/langchain/functions/request-handler/adapters/bedrock/base.py. Github hides this file by default
- since all the models use the same prompts. LLama custom prompts are no longer used and it relies on the Converse API.

**Testing**
Called the models one by and checked the usage 

image


lib/model-interfaces/langchain/functions/request-handler/adapters/base/base.py
```diff
@@ -71,12 +71,18 @@ def on_llm_end(
 
 class ModelAdapter:
     def __init__(
-        self, session_id, user_id, mode=ChatbotMode.CHAIN.value, model_kwargs={}
+        self,
+        session_id,
+        user_id,
+        mode=ChatbotMode.CHAIN.value,
+        disable_streaing=False,
```
bigadsoleiman, line 78: typo `_streaing`

lib/model-interfaces/langchain/functions/request-handler/adapters/base/base.py
```diff
@@ -176,12 +182,14 @@ def run_with_chain_v2(self, user_prompt, workspace_id=None):
 
         config = {"configurable": {"session_id": self.session_id}}
         try:
-            if self.model_kwargs.get("streaming", False):
+            if self.disable_streaming == False and self.model_kwargs.get(
```
bigadsoleiman, line 185: shorter `if not self.disable_streaming`

bigadsoleiman (approved):

PR #573, opened 2024-09-20
feat: Add option to use CMK and set the resources retention

charles-marion:
*Issue #, if available:*

*Description of changes:*
* Add an option to use a customer managed keys for data encryption.
* Add an option to retain data stores and logs on cleanup
* Add versioning to all buckets

*Testing*
* Verified the generated template without change of configuration has no breaking change
* Migrated an existing environment to use CMK. The difference can be seen [in this commit](https://github.com/charles-marion/aws-genai-llm-chatbot/commit/12d0701517634c839fea880a598e90ef8130d6e7#diff-57c864f4d012e0ddefb56955c19f6e4e4722e274ea9e5c2f801669b69ad6e13fL321) (/tests/__snapshots__/cdk-app.test.ts.snap)
* Manual tests + Integration tests


lib/sagemaker-model/index.ts
```diff
@@ -17,15 +15,13 @@ export class SageMakerModel extends Construct {
     const { model } = props;
     this.modelId = model.modelId;
 
-    if (model.type == DeploymentType.Container) {
-      const { endpoint } = deployContainerModel(this, props, model);
```
charles-marion, line 21: Removed because it was never used.

lib/rag-engines/aurora-pgvector/index.ts
```diff
@@ -30,7 +30,13 @@ export class AuroraPgVector extends Construct {
       engine: rds.DatabaseClusterEngine.auroraPostgres({
         version: rds.AuroraPostgresEngineVersion.VER_15_3,
       }),
-      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      storageEncryptionKey: props.shared.kmsKey,
+      // Always setting it to true would be a breaking change.
+      storageEncrypted: props.shared.kmsKey ? true : false,
```
charles-marion, line 35: My preference would be to always enable it but it would trigger the recreation of the table

lib/chatbot-api/chatbot-dynamodb-tables/index.ts
```diff
@@ -19,8 +25,14 @@ export class ChatBotDynamoDBTables extends Construct {
         type: dynamodb.AttributeType.STRING,
       },
       billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
-      encryption: dynamodb.TableEncryption.AWS_MANAGED,
-      removalPolicy: cdk.RemovalPolicy.DESTROY,
+      encryption: props.kmsKey
+        ? dynamodb.TableEncryption.CUSTOMER_MANAGED
+        : dynamodb.TableEncryption.AWS_MANAGED,
+      encryptionKey: props.kmsKey,
```
grinko, line 31: What happens if props.kmsKey does not exist?

lib/rag-engines/aurora-pgvector/index.ts
```diff
@@ -73,7 +79,8 @@ export class AuroraPgVector extends Construct {
 
     const dbSetupResource = new cdk.CustomResource(
       this,
-      "DatabaseSetupResource",
+      // Force recreation on CMK cange to re-init the DB cluster.
```
grinko, line 82: typo: cange

lib/chatbot-api/chatbot-s3-buckets/index.ts
```diff
@@ -1,29 +1,48 @@
...
       blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
-      removalPolicy: cdk.RemovalPolicy.DESTROY,
-      autoDeleteObjects: true,
+      removalPolicy:
+        props.retainOnDelete === true
+          ? cdk.RemovalPolicy.RETAIN_ON_UPDATE_OR_DELETE
+          : cdk.RemovalPolicy.DESTROY,
+      autoDeleteObjects: props.retainOnDelete !== true,
       enforceSSL: true,
+      encryption: s3.BucketEncryption.S3_MANAGED,
```
grinko, line 27: Why does logs bucket encrypted with S3 managed key and files bucket has options?

lib/chatbot-api/chatbot-dynamodb-tables/index.ts
charles-marion: It passes undefined. (property is ignored)

lib/chatbot-api/chatbot-s3-buckets/index.ts
charles-marion: It is not supported.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html

"You can use [default bucket encryption](https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-bucket-encryption.html) on the destination bucket only if you use server-side encryption with Amazon S3 managed keys (SSE-S3), which uses the 256-bit Advanced Encryption Standard (AES-256). Default server-side encryption with AWS Key Management Service (AWS KMS) keys (SSE-KMS) is not supported."

lib/rag-engines/aurora-pgvector/index.ts
charles-marion:
```suggestion
      // Force recreation on CMK change to re-init the DB cluster.
```

grinko (approved):

bigadsoleiman (approved):

Issue #235, opened 2023-11-22
old version of pgvector extension


alexeyshishkin01:
Hello, all !

Could you pls. update the version of pgvector extension ?
My setup shows the following:

psql --host=XXX.us-east-1.rds.amazonaws.com --user=postgres
Password for user postgres:
psql (16.1 (Ubuntu 16.1-1.pgdg22.04+1), server 15.3)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
Type "help" for help.

postgres=> \dx
                        List of installed extensions
  Name   | Version |   Schema   |                Description
---------+---------+------------+--------------------------------------------
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
 vector  | 0.4.1   | public     | vector data type and ivfflat access method
(2 rows)

The latest version of pgvector now is 0.5.1 and it's highly recommended to use it due to the fixes it has.

PR #578, opened 2024-09-25
feat: Use IAM with RDS, increase number of backups and upgraded engine.

charles-marion:
*Issue #, if available:*
#235
*Description of changes:*

Aurora changes:
* Upgrade engine (15.4->15.7) and related pgvector extension (0.4.1 -> 0.7)
* Add backups configuration (35 days)
* Add DB Users: Admin, Read, Write. Use these users for IAM based auth instead of password/secret based.

Note this change might lead to a small downtime of the Aurora DB (I did not notice it)
Also note the password is still use for the initial setup of the users.

Testing:
* Manually ran a search before and after the upgrade. Verified the vector scores were the same.
* Ran integration tests.
* Verified all step functions were passing.


lib/rag-engines/aurora-pgvector/index.ts
```diff
@@ -73,7 +86,7 @@ export class AuroraPgVector extends Construct {
 
     const dbSetupResource = new cdk.CustomResource(
       this,
-      "DatabaseSetupResource",
+      "DatabaseSetupExtensionsAndUsers",
```
charles-marion, line 89: Rename to force running the script on existing environments.

grinko (approved):

PR #581, opened 2024-10-03
feat: Added rate limit + security headers.

charles-marion:
*Issue #, if available:*

*Description of changes:*
* Upgrade the cloudfront CDK Construct. `CloudFrontWebDistribution` is deprecated
* Add [Security headers](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-responseheaderspolicy-securityheadersconfig.html)
* Added WAF to throttled the Appsync request per IP. Added one rule specific for `sendQuery`

*Testing*
* Verified throttling from both cloudfront and appsync by call many time the LLM
* Verified the front end still works with federation, image upload. The content `ContentSecurityPolicy` header is strict to prevent using unexpected resources.

### DIFF
--- a/lib/chatbot-api/index.ts
+++ b/lib/chatbot-api/index.ts
@@ -5,6 +5,7 @@ import * as sqs from "aws-cdk-lib/aws-sqs";
 import * as sns from "aws-cdk-lib/aws-sns";
 import * as ssm from "aws-cdk-lib/aws-ssm";
 import * as iam from "aws-cdk-lib/aws-iam";
+import * as wafv2 from "aws-cdk-lib/aws-wafv2";
 import * as cdk from "aws-cdk-lib";
 import * as path from "path";
 import { Construct } from "constructs";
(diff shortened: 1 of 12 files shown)


lib/chatbot-api/index.ts
```diff
@@ -152,4 +174,74 @@ export class ChatBotApi extends Construct {
       },
     ]);
   }
+
+  private createWafRules(llmRatePerIP: number): wafv2.CfnWebACL.RuleProperty[] {
+    /**
+     * The rate limit is the maximum number of requests from a
+     * single IP address that are allowed in a five-minute period.
```
bigadsoleiman, line 181: is it five or ten minute period?

bigadsoleiman (approved):

lib/chatbot-api/index.ts
charles-marion, line 181:
```suggestion
     * single IP address that are allowed in a ten-minute period.
```

lib/shared/index.ts
```diff
@@ -316,4 +320,32 @@ export class Shared extends Construct {
       { id: "AwsSolutions-SMG4", reason: "Secret value is blank." },
     ]);
   }
+
+  private createWafRules(ratePerIP: number): wafv2.CfnWebACL.RuleProperty[] {
+    /**
+     * The rate limit is the maximum number of requests from a
+     * single IP address that are allowed in a five-minute period.
```
charles-marion, line 327:
```suggestion
     * single IP address that are allowed in a ten-minute period.
```

Issue #580, opened 2024-09-26
Enabling RAG makes it relatively easy to reach 400KB limit for a dynamodb item on sessions table


gbone-restore:
Amazon DynamoDB limits the size of each item in a table to 400 KB. It's feasible that this limit could be reached on the sessions table because of all the extra metadata stored in the item. Especially when running with RAG configured. The session table stores chat history in a single item attribute.

To work around this limitation, we can store chat history as individual items in the sessions table. I can submit a PR for this change if there is an interest in this issue.

charles-marion:
Hi @gbone-restore ,

I agree it is an issue but changing the table structure might be a significant change.

A shorter/smaller change could be to reduce the metadata stored and limit the number of message in the sessions.

gbone-restore:
Here's the data model that I used to test out this change.

![Sessions](https://github.com/user-attachments/assets/4fe47f50-1bb9-4e25-b569-27d644a7d120)

The partition key is now users and the sort key is sessions. And there are two types of items to make selecting session history and dealing with chat history a little easier. I can submit a PR with a working example. The bulk of the changes are in `lib/shared/layers/python-sdk/python/genai_core/langchain/chat_message_history.py` and `lib/shared/layers/python-sdk/python/genai_core/sessions.py`.

charles-marion:
> The partition key is now users and the sort key is sessions.
I think it's a good solution. My concern is it would invalidate existing sessions

> I can submit a PR with a working example
If you already have a prototype available it would be welcome.

Note I intend to discuss this topic with the maintainers.

PR #584, opened 2024-10-11
New sessions table that allows for chat history and metadata greater than 400KB

gbone-restore:
Issue #, if available:
https://github.com/aws-samples/aws-genai-llm-chatbot/issues/580

Description of changes:
Session History table was redesigned to handle chat history and meta data that is greater than 400KB. 
The partition key is now users and the sort key is sessions. And there are two types of items to make selecting session history and dealing with chat history a little easier. I can submit a PR with a working example. The bulk of the changes are in lib/shared/layers/python-sdk/python/genai_core/langchain/chat_message_history.py and lib/shared/layers/python-sdk/python/genai_core/sessions.py.

### DIFF
--- a/lib/aws-genai-llm-chatbot-stack.ts
+++ b/lib/aws-genai-llm-chatbot-stack.ts
@@ -77,7 +77,6 @@ export class AwsGenAILLMChatbotStack extends cdk.Stack {
         ragEngines,
         messagesTopic: chatBotApi.messagesTopic,
         sessionsTable: chatBotApi.sessionsTable,
-        byUserIdIndex: chatBotApi.byUserIdIndex,
       });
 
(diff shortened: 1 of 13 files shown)


gbone-restore:
I will take a look at the failed tests. I made this change prior to the tests being available, so there probably is a little cleanup needed to get the tests to pass.

charles-marion:
Hi @gbone-restore , Thank you for creating this PR.

I reviewed the change and I have minor suggestions (not published)
However, this change would break existing deployment of the project since the change or partition key would invalidate existing sessions. 

A possible option is to apply a similar change with a new table for messages.

We will follow up with a new PR based on your change.

gbone-restore:
Agreed. That is how I initially implemented things inside my environment. This would be an easy change and I can update this PR to create a new sessions table.

Issue #583, opened 2024-10-10
Lambda was unable to configure your environment variables because the environment variables you have provided exceeded the 4KB limit.


jarrettyeo:
```
12:46:25 AM | UPDATE_FAILED        | AWS::Lambda::Function                       | ChatBotApiRestApiG...ApiHandler140797C1
1 validation error detected: Value 'Lambda was unable to configure your environment variables because the environment variables you have provided exceeded the 4KB limit.
```

After the commit of 05f2b75611318d7978e781f19d2cd9badde08dae https://github.com/aws-samples/aws-genai-llm-chatbot/pull/578, I started to encounter this error when I deploy CDK

jarrettyeo:
Might need to start using Parameter Store to store and access these env vars

jarrettyeo:
Likely related to https://github.com/aws-samples/aws-genai-llm-chatbot/commit/05f2b75611318d7978e781f19d2cd9badde08dae#diff-07e07c5729f4979449d5393179aafa12da09ae153006a131d371faa58d674a60R79 @charles-marion

charles-marion:
Hi @jarrettyeo , thank you for noticing the issue.
Please use the previous commit if you are blocked. I will look into it.

PR #589, opened 2024-10-11
feat: Migrate several env variables to SSM to prevent reaching lambda…

charles-marion:
…'s limit.

*Issue #, if available:* 
#583

*Description of changes:*
* Use the Lambda extension to load SSM Parameters
* Replace large env variable (ARN/DNS) to use this solution

If the issue #583 is still happening after merging, we can add move environment variables to the list.

*Testing*
Manual and integration tests after deployment.


charles-marion:
Closing, this change it not required. Could re-opened is other environments are impacted.

Issue #596, opened 2024-10-28
Error while deploying the stack


surukonda:
Trying to deploy the stack causes deployment to fail with following error.

```
2:07:02 PM | CREATE_FAILED        | AWS::StepFunctions::StateMachine                | RagEnginesAuroraPg...aWorkspace50EFF4E7
Resource handler returned message: "Invalid Logging Configuration: The CloudWatch Logs Resource Policy size was exceeded. We suggest prefixing your CloudWatch log group name with /aws/vendedlogs/states/. (Service: AWSStepFunctions; Status Code: 400; Error Code: Inva
lidLoggingConfiguration; Request ID: c2ca8c66-348b-4413-88bb-80d913450642; Proxy: null)" (RequestToken: 2ab965eb-f585-b0b1-8da9-b06b54ff745d, HandlerErrorCode: InvalidRequest)
```
Here is my config.json file contents:
```json
{
  "prefix": "proto",
  "createCMKs": false,
  "retainOnDelete": false,
  "bedrock": {
    "enabled": true,
    "region": "us-east-1",
    "guardrails": {
      "enabled": false,
      "identifier": "",
      "version": "DRAFT"
    }
  },
  "llms": {
    "enableSagemakerModels": false,
    "sagemaker": [],
    "huggingfaceApiSecretArn": ""
  },
  "rag": {
    "enabled": true,
    "deployDefaultSagemakerModels": false,
    "engines": {
      "aurora": {
        "enabled": true
      },
      "opensearch": {
        "enabled": true
      },
      "kendra": {
        "enabled": false,
        "createIndex": false,
        "external": [],
        "enterprise": false
      },
      "knowledgeBase": {
        "enabled": false,
        "external": []
      }
    },
    "embeddingsModels": [
      {
        "provider": "bedrock",
        "name": "amazon.titan-embed-text-v1",
        "dimensions": 1536,
        "default": false
      },
      {
        "provider": "bedrock",
        "name": "amazon.titan-embed-image-v1",
        "dimensions": 1024,
        "default": false
      },
      {
        "provider": "bedrock",
        "name": "cohere.embed-english-v3",
        "dimensions": 1024,
        "default": false
      },
      {
        "provider": "bedrock",
        "name": "cohere.embed-multilingual-v3",
        "dimensions": 1024,
        "default": true
      },
      {
        "provider": "openai",
        "name": "text-embedding-ada-002",
        "dimensions": 1536,
        "default": false
      }
[...]

Issue #592, opened 2024-10-22
npm run config fails reproducible (locally + Cloud9) with default settings


markusz:
Running `npm run config` fails reproducible when following the [deploy](https://aws-samples.github.io/aws-genai-llm-chatbot/guide/deploy.html#deployment) instructions. 

I get the following error:

```
Could not complete the operation.
Cannot set properties of undefined (setting 'default')
```

I verified `npm run config` failing in different environments and configurations
- node v18 & v20
- Cloud9 & locally
- Default config values vs custom config

### Steps to reproduce:

1. Clone the repository and checkout `main`. This will point to `v3.1.3` of the aws-genai-llm-chatbot 
2. Follow the instructions for Cloud9 / local deployment
3. Use default values for all inputs

![image](https://github.com/user-attachments/assets/d7ed3ffe-a182-4a19-9f20-6b3f716ad5da)

markusz:
v4.0.14 works, but `main` still points to `v3.1.3` so this likely affects everyone following the deploy instructions in the docs

markusz:
### Workaround

- Download `v4.0.14` or higher manually from [here](https://github.com/aws-samples/aws-genai-llm-chatbot/tags)
- Unzip
- Continue with [Step 2](https://aws-samples.github.io/aws-genai-llm-chatbot/guide/deploy.html#deployment) of the deployment

charles-marion:
Added a fix in #594

markusz:
That was fast. Thanks!

yoavchaws:
Well, when I run "npm run config" in Cloud9 in eu-central-1, selecting RAG with Knowledge base, it exits without asking me to save a config file and no config.json is created. Same with "npm run create". I suspect it is related to the change done to config.ts for this fix.
![2024-10-24_09-41-02](https://github.com/user-attachments/assets/427607ed-497f-4e07-85a7-7fedd58d5103)

yoavchaws:
Thanks @charles-marion. The fix works fine. However, when I create a bedrock based RAG configuration, it does not ask me for the embedding model (embeddingModels[]) to use. As a result, the cdk bootstrap operation falls later with "no embedding model set". 
A **work around** is to change one of the embeddingModels[].default to true.

charles-marion:
Added a fix in the PR https://github.com/aws-samples/aws-genai-llm-chatbot/pull/598 .
Please use the work around listed above until merged.

charles-marion:
[...]

PR #598, opened 2024-10-29
bug: Upgrade Langchain, AWS Powertools, Pydantic. Fix config without default embedding and step function deployment.

charles-marion:
*Issue #, if available:* #596

*Description of changes:*

In #596 , the Step function prevents deployment because the account has reach the number of policies. See here
https://docs.aws.amazon.com/step-functions/latest/dg/sfn-best-practices.html#bp-cwl

The fix is to set the prefix of the log group names (`/aws/vendedlogs/`) so step function uses statement for all..

In #592, the CDK synth fails when there is no default embedding model (only needed when using Aurora/Opensearch).
Removed the check (since it's a valid use case) and verified the Embedding page is still working without default model. 

*Dependencies updates:*
Because the build workflow fails due to a vulnerabilities in Langchain (https://github.com/aws-samples/aws-genai-llm-chatbot/security/dependabot/73), I upgraded the following

* Langchain: [Guide](https://python.langchain.com/docs/versions/v0_3/)
* Because it conflicts with Langchain, I also had to upgrade  [AWS Power tools](https://docs.powertools.aws.dev/lambda/python/3.2.0/upgrade/)
* Because both the above require a new version of pydantic, I also upgraded it.  In this case, the project uses it mostly for validator which has a new bahevior for optional values. ([Guide](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields))

*Testing*
Manual testing + Integration test.


grinko (approved):

surukonda:
@charles-marion please make this change in `lib/rag-engines/workspaces/delete-document.ts` as well.

charles-marion:
> @charles-marion please make this change in `lib/rag-engines/workspaces/delete-document.ts` as well.

done

grinko (approved):

PR #603, opened 2024-11-07
fix s3 issue if deploying from region other than us-east-1

arief-hidayat:
*Issue #, if available:*
encounter S3 issue when deploying from ap-southeast-3 (other than us-east-1), such as CORS. 

*Description of changes:*
follow recommendation from https://aws.amazon.com/blogs/media/deep-dive-into-cors-configs-on-aws-s3-how-to/

### DIFF
--- a/lib/rag-engines/data-import/functions/upload-handler/index.py
+++ b/lib/rag-engines/data-import/functions/upload-handler/index.py
@@ -7,12 +7,13 @@ import genai_core.workspaces
 from aws_lambda_powertools import Logger, Tracer
 from aws_lambda_powertools.utilities.data_classes import SQSEvent, event_source
 from aws_lambda_powertools.utilities.typing import LambdaContext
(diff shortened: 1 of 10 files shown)


charles-marion:
Hi @arief-hidayat ,

Thank you for the update! (and sorry for the late follow up)
Can you clarify what is the issue in other regions (or error you got)?

The use of `s3v4` was already set for all pre-signed URLs (which is when CORS would matter). (And the SDK automatically set the current region)

That being said I don't see a downside of this change since v4 is the current version

arief-hidayat:
If I remember correctly, there is CORS issue because of the behaviour mentioned in the blog (about redirect URL). So, I followed the blog recommendation and it worked (at least tested on ap-southeast-3 case).

I haven't tried removing `s3v4`, perhaps it will still work without it. But without setting region, it failed.

PR #611, opened 2024-11-30
Fix/model name validation

gbone24:
*Issue #, if available:*

*Description of changes:*
constants used during validation do not include a colon `:`.  This breaks validating titan v2 embedding models that actually have a colon in their name: "amazon.titan-embed-text-v2:0"
image


charles-marion:
Thank you for the change. Adding support for Titan v2 is a great addition.

There are style issues (npm run vet-all)

If you have time please have a look otherwise I can do it at a later time.

PR #613, opened 2024-12-03
Amazon Nova Models Support

bigadsoleiman:
*Issue #, if available:*

*Description of changes:*
Adding support of new Amazon Nova models
- `Amazon Nova Micro`
- `Amazon Nova Lite`
- `Amazon Nova Pro`
- `Amazon Nova Canvas`
- `Amazon Nova Reels`

**Backend**
- Prevented the layer builder to remove boto3\* botocore\* since we need to install them from `.whl` this is until new boto3 apis are made public
- Added to langchain adapter a new regex to register `bedrock.amazon.nova` to use existing `BedrockChatAdapter` this will only work for `Nova Micro` since the other has to go through multimodal adapter.
- Changed `SAFE_STR_REGEX` to support multi-path s3 paths (i.e. private/user_id/videos/..../\*.mp4) since the video generations forcefully puts videos under "randomid/output.mp4" under any S3 URI passed as output path.
- Changed `MultiModalBase` to support file upload/retrieval to/from S3, converse API and generation of non-text content (video/image)
- Adapted `Claude`/ `Idefics` adapters to new MultiModalBase class by returning the expected new structure.
- Created `Nova` adapter from MultiModalBase (using Converse API where currently possible)
  * Video generation relies on a new API `bedrock.start_async_invoke` which generates 6s video in background asynchronously in \~4m.

**UI**

- `chat-message.tsx` has been broken down in smaller components.
- added a `chat-message-media-display.tsx` to handle the display of media (image/video)
  * For video generation it assumes that if the S3 signed url is not avaible the video is still generating and does a long polling (every 30seconds) until the video is availble. This is a temporary solution to support async generation. A more robust backend workflow (stepfunctions?) should be implemented for long term.
- added `video-dialog` for video uploads (max 10MB)
- added `seed` options defaulting to 0 in `llm-config-dialog`, used only in image/video generation.
- added support for single selection of multiple OUTPUT modalities in the `chat-input-panel.tsx` based on model's OUTPUT_MODALITIES, default to first or only available one.

Ran `vet-all` for linting and testing.


charles-marion (approved):
Thank you for the change!.

lib/shared/web-crawler-batch-job/requirements.txt
```diff
@@ -4,7 +4,7 @@ numpy==1.26.0
 cfnresponse==1.1.2
 aws_requests_auth==0.4.3
 requests-aws4auth==1.2.3
-langchain==0.3.7
```
charles-marion, line 7: I would recommend to revert.

lib/shared/layers/common/requirements.txt
```diff
@@ -1,12 +1,14 @@
+boto3-1.35.58-py3-none-any.whl
+botocore-1.35.58-py3-none-any.whl
+langchain_aws-0.2.2-py3-none-any.whl
 aws_xray_sdk==2.14.0
 numpy==1.26.0
 cfnresponse==1.1.2
 aws_requests_auth==0.4.3
 requests-aws4auth==1.2.3
-langchain==0.3.7
```
charles-marion, line 6: I would recommend to revert.

PR #621, opened 2025-01-07
feat: Main branch validation

charles-marion:
*Issue #, if available:*

*Description of changes:*
Trigger validation on changes on the main branch.

### DIFF
--- a/.github/workflows/e2e-validation.yml
+++ b/.github/workflows/e2e-validation.yml
@@ -15,13 +15,14 @@ jobs:
       uses: aws-actions/configure-aws-credentials@v2
       with:
         aws-region: ${{ secrets.PIPELINE_AWS_REGION }}
-        role-to-assume: arn:aws:iam::${{ secrets.PIPELINE_AWS_ACCOUNT_ID }}:role/${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
+        role-to-assume: arn:aws:iam::${{ secrets.PIPELINE_AWS_ACCOUNT_ID }}:role/${{ secrets.PIPELINE_ROLE }}
         role-session-name: github-action
     - name: Validate
       run: |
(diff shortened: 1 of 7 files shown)


integtests/chatbot-api/aurora_workspace_test.py
```diff
@@ -84,7 +84,7 @@ def test_add_rss(client: AppSyncClient):
 
     ready = False
     retries = 0
-    while not ready and retries < 10:
+    while not ready and retries < 20:
```
maryamkhidir, line 87: nit: Why are we increasing the retries here?

.github/workflows/e2e-validation.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: aws-actions/configure-aws-credentials@v2
+      with:
+        aws-region: ${{ secrets.PIPELINE_AWS_REGION }}
+        role-to-assume: arn:aws:iam::${{ secrets.PIPELINE_AWS_ACCOUNT_ID }}:role/${{ secrets.PIPELINE_ROLE }}
+        role-session-name: github-action
+    - name: Validate
+      run: |
+          executionId=$(aws codepipeline start-pipeline-execution --name  ${{ secrets.PIPELINE_NAME }} --region ${{ secrets.PIPELINE_AWS_REGION }} --query "pipelineExecutionId" --output text)
+          sleep 60
+          while true; do
```
maryamkhidir, line 23: Do we need to add a timeout here for indefinite runs or this is already handled by aws code pipeline?

integtests/chatbot-api/aurora_workspace_test.py
charles-marion: In some case the test was slow/flaky. I increased the wait time to see if it improves it.

.github/workflows/e2e-validation.yml
charles-marion: There is a default one of 6h. We could set one if needed.

PR #624, opened 2025-01-22
feat: Application configuration

maryamkhidir:
### Breaking Changes
To use the project, Coginto user need a Cognito group (for example admin, user). Chatbot design was replaced to use CloudScape design pattern.

### New Features
- Support Amazon Nova LLMs Integration.
- Add bedrock guardrails to all models in the langchain interface.
- Add document support to chatbot sessions.
- Add custom system prompts for non-bedrock models.
- Add happy path selenium tests.
- Update chatbot to use latest cloudscape components.
- Create configurable applications with frontend implementation.

### Fixes & Improvements
- Bug fixes:
	- Fixed video understanding input.
	- Fixed integration tests following appsec findings.
	- Fixed input validation & presign URL role verification.
	- Fixed federation + application update.
	- Fixed RSS Feed, Application validation, RDS permissions.
	- Fixed semantic search validation.
	- Fixed import and unit test snapshots.
- Security improvements:
	- Disabled Appsync introspection.
	- Updated security documentation.
- Access control improvements:
	- Added access control to APIs.
	- Added unauthorized unit tests.
	- Implemented sendQuery access control.
	- Added role API.
	- Hide application attributes for non-admin users.

*Issue #, if available:*

*Description of changes:*


charles-marion (approved):
Approved. All integration tests are passing.

azaylamba:
@maryamkhidir Very good feature and a lot of changes, thanks for this.
I have few suggestions to make it more useful and practical:
1. Applications should have access to user history. Potentially can include the navigation panel (with restricted menus) in the `application.tsx`component.
2. Once the history is accessible, the user should be able to continue the conversion. I see a problem there at the moment because the application chat uses route `http://localhost:3000/application/<application-id>` while the session in the session history uses `http://localhost:3000/chatbot/playground/<session-id>` which is different route and the application user might not have access to that route. So it is tricky to allow continued conversations. 
Should we add another route for application chat history and continued conversations?

azaylamba:
@maryamkhidir It seems the user feedback feature is not working after these changes because message history item doesn't have metadata now. The metadata is attached only if the user has admin or workspace_manager role.

azaylamba:
@maryamkhidir @charles-marion Also I think we should use applicationId in sessions table and fetch the session history based on the applicationId.

Issue #609, opened 2024-11-25
Feature Request - Support for Inference profiles in Bedrock


azaylamba:
Some of the models available on Bedrock e.g. Claude 3.5 Sonnet v2 can be used through an inference profile only. We should support for creating and using an inference profile so that these newer models can be used.

PR #629, opened 2025-02-02
Feature: Support Cross Region Inference

michaelraskansky:
Issue #, if available:
https://github.com/aws-samples/aws-genai-llm-chatbot/issues/609

*Description of changes:*
This change enables the **listModels** API to return models that support cross-region inference and makes necessary changes to allow model providers to register themselves to handle cross-region inference calls. Additionally, the Lambda function permissions have been modified to allow listing of inference profiles.


lib/shared/layers/python-sdk/python/genai_core/models.py
```diff
@@ -80,6 +84,73 @@ def list_azure_openai_models():
...
+        ]
+        for inference_profile in response.get("inferenceProfileSummaries", [])
+        if (
+            inference_profile.get("status") == "ACTIVE"
+            and inference_profile.get("type") == "SYSTEM_DEFINED"
+        )
+    }
+
+
+def list_bedrock_cris_models():
```
maryamkhidir, line 137: You mean `list_bedrock_cross_models`?

michaelraskansky: cris is short for cross region inference profiles

maryamkhidir (approved):

lib/shared/layers/python-sdk/python/genai_core/models.py
maryamkhidir: Approved, thank you!

michaelraskansky: Sure, thanks!

PR #653, opened 2025-04-21
Add additional Bedrock Knowedlge Base supported regions

sleepwithcoffee:
*Issue #, if available:*

#23

*Description of changes:*

### DIFF
--- a/cli/magic-config.ts
+++ b/cli/magic-config.ts
@@ -11,6 +11,7 @@ import {
   SystemConfig,
   SupportedBedrockRegion,
   ModelConfig,
+  SupportedBedrockKnowledgeBaseRegion,
 } from "../lib/shared/types";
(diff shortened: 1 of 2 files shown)


maryamkhidir:
can you add a screenshot of this change in the cli?

sleepwithcoffee:
Image
Image

sleepwithcoffee:
hi @Rob-Powell this is Bedrock Knowledge Base region, isn't it different and should not impact in any way?

Rob-Powell:
> hi @Rob-Powell this is Bedrock Knowledge Base region, isn't it different and should not impact in any way?

no, as batch is how the documents get into the knowledge base via the solution. I don't think it makes sense to enable more knowledge base bedrock regions, if we know they wont work for core use cases of the solution.

PR #663, opened 2025-05-20
fix: update vite to fix security vulnerability

cajias:
This PR updates the vite package to fix the moderate security vulnerability (GHSA-859w-5945-r5v3) related to server.fs.deny being bypassed with /. for files under project root.


maryamkhidir (approved):
Thanks for working on this

PR #673, opened 2025-07-30
fix: add custom attribute to cognito for OIDC and SAML deployments

maryamkhidir:
*Issue #, if available:*

### Description
This change adds a custom attribute configuration to the Cognito User Pool to support OIDC (OpenID Connect) and SAML authentication deployments. This enhancement improves the authentication flow by ensuring proper attribute mapping for federated identity providers.

### Files Changed
- **lib/authentication/index.ts** - Core authentication logic updated to include custom attribute configuration
- **Test snapshots updated** to reflect the new Cognito configuration:
    • tests/__snapshots__/cdk-app.test.ts.snap
     • tests/authentication/__snapshots__/autentication-construct.test.ts.snap
     • tests/chatbot-api/__snapshots__/chatbot-api-construct.test.ts.snap

### Impact
• Enables proper attribute handling for federated authentication scenarios
• Maintains backward compatibility with existing authentication flows
• Updates test snapshots to ensure CI/CD pipeline continues to pass

### Testing
• All existing tests updated with new snapshots


lib/authentication/index.ts
```diff
@@ -34,6 +34,9 @@ export class Authentication extends Construct {
       signInAliases: {
         email: true,
       },
+      customAttributes: {
```
charles-marion, line 37: I believe this is only used when federation is used. I would conditionally set this.

To my knowledge the attribute was set without this cdk property. (but I only tested once a while back)

Also note this other change: https://github.com/aws-samples/aws-genai-llm-chatbot/pull/672/files

charles-marion (approved):
Approved but please review my comment before merging.

lib/authentication/index.ts
maryamkhidir: Done, updated in https://github.com/aws-samples/aws-genai-llm-chatbot/pull/673/commits/3365883c5b2b57aeb0d2458985e45deb19db5611

PR #726, opened 2026-03-11
feat: allow AgentCore agent selection in Applications

michaelraskansky:
## Summary

Adds the ability for admins to configure an Application with an Amazon Bedrock AgentCore agent, optionally alongside a model and model settings. Regular users can then chat with agents through the Application interface.

### Backend
- **GraphQL schema**: `agentRuntimeArn` field on `Application` type and `ManageApplicationInput`, `model` made nullable
- **genai_core/applications.py**: agent ARN validation (strict regex), model-or-agent-required logic in `validate_request`, agent ARN stored/returned in CRUD operations
- **routes/applications.py**: Pydantic models updated with `Optional[str]` model and `agentRuntimeArn` with ARN pattern validation
- **send-query-lambda-resolver**: routes agent-based applications to `modelInterface: "agent"` (set server-side), model-based to `langchain`. Passes `modelName`, `provider`, and `modelKwargs` to agent when configured
- **Bug fix**: `ConsiseSystemPrompt` → `CondenseSystemPrompt` typo in `update_application`

### Frontend
- Agent selector in application form (filters by `READY` status)
- Model and model settings remain available when agent is selected (agent receives them in payload)
- Dynamic `modelInterface` in chat-input-panel based on application config

### Testing
- **24 unit tests** (17 route-level + 7 validate_request including ARN injection attempt)
- **10 integration tests** (6 existing model-based + 4 agent: create, get, query with content verification, delete)

## How it works

1. Admin creates an Application, selects an AgentCore agent (and optionally a model + settings)
2. When a user sends a message through the Application, the send-query resolver sets `modelInterface: "agent"` server-side
3. SNS routes the message to the Bedrock Agents SQS queue
4. The Bedrock Agents Lambda invokes the agent via AgentCore with the full payload (including model info and kwargs)
5. Agent streams SSE responses back through the existing WebSocket pipeline

### DIFF
--- a/integtests/chatbot-api/application_test.py
+++ b/integtests/chatbot-api/application_test.py
@@ -134,3 +134,81 @@ def test_update_application(client: AppSyncClient):
 def test_delete_application(client: AppSyncClient):
     result = client.delete_application(pytest.application.get("id"))
     assert result == True
+
+
+# --- Agent application integration test ---
+
+AGENT_RUNTIME_ARN = "arn:aws:bedrock-agentcore:eu-central-1:481665129290:runtime/rapidRed_Agent-0S40yv5Brf"
+
+
+def test_create_agent_application(client: AppSyncClient):
+    pytest.agent_application = client.create_application(
+        input={
+            "name": "INTEG_TEST_AGENT_APP",
+            "agentRuntimeArn": AGENT_RUNTIME_ARN,
+            "roles": ["user"],
+            "allowImageInput": False,
+            "allowVideoInput": False,
+            "allowDocumentInput": False,
+            "enableGuardrails": False,
+            "streaming": True,
+            "maxTokens": 512,
+            "temperature": 0.6,
+            "topP": 0.9,
+        }
+    )
+
+    assert pytest.agent_application.get("id") is not None
+    assert pytest.agent_application.get("name") == "INTEG_TEST_AGENT_APP"
+    assert pytest.agent_application.get("agentRuntimeArn") == AGENT_RUNTIME_ARN
+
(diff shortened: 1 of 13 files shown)


JiwaniZakir (commented):
The hardcoded `AGENT_RUNTIME_ARN` in `application_test.py` (line `AGENT_RUNTIME_ARN = "arn:aws:bedrock-agentcore:eu-central-1:481665129290:runtime/..."`) contains what appears to be a real AWS account ID in a public repository — this should be replaced with an environment variable or a placeholder value pulled from test configuration, similar to how other test fixtures handle account-specific resources.

In `applications.py`, both `model` and `agentRuntimeArn` are now `Optional` on `CreateApplicationRequest` and `UpdateApplicationRequest`, but there's no cross-field validator (e.g., Pydantic's `@model_validator`) enforcing that exactly one of the two is provided. As written, a request with neither field set would pass validation and likely cause a confusing downstream failure, and a request with both set simultaneously has undefined behavior.

The `ARN_REGEX = r"^[A-Za-z0-9-_.:/]+$"` is quite permissive — it would accept arbitrary strings like `abc:123` that are structurally nothing like an ARN. Consider anchoring it to the expected Bedrock AgentCore ARN prefix (e.g., `^arn:aws:bedrock-agentcore:[a-z0-9-]+:\d{12}:runtime/[\w-]+$`) so validation catches malformed values early at the API boundary rather than letting them propagate to DynamoDB or downstream calls.

maryamkhidir:
@michaelraskansky please rebase and fix the build issues

PR #727, opened 2026-03-24
feat: add MiniMax as first-class LLM provider

octo-patch:
## Summary

This PR adds [MiniMax](https://www.minimax.io/) as a first-class third-party LLM provider, following the same adapter-registry pattern used by OpenAI and Azure OpenAI.

**Supported models:**
| Model | Context | Description |
|-------|---------|-------------|
| MiniMax-M2.7 | 1M tokens | Latest flagship model |
| MiniMax-M2.5 | 204K tokens | Strong reasoning |
| MiniMax-M2.5-highspeed | 204K tokens | Faster variant |

**Key changes (11 files, 421 additions):**
- `genai_core/types.py`: Add `MINIMAX` to Provider enum
- `genai_core/clients.py`: Add `get_minimax_client()` using OpenAI SDK with custom base URL
- `genai_core/model_providers/direct/provider.py`: Add `_list_minimax_models()` for auto-discovery
- `adapters/minimax/minimax_chat.py`: `MinimaxChatAdapter` extending `ModelAdapter` via `ChatOpenAI` with OpenAI-compatible API
- `adapters/__init__.py`: Register minimax adapter module
- `docs/documentation/model-requirements.md`: MiniMax setup documentation

**How it works:**
- MiniMax uses an OpenAI-compatible API at `https://api.minimax.io/v1`
- The adapter uses LangChain's `ChatOpenAI` with a custom `openai_api_base`
- Temperature is clamped to `[0, 1]` per MiniMax API requirements
- Models auto-appear in the UI when `MINIMAX_API_KEY` is added to Secrets Manager

## Test plan

- [x] 11 unit tests for `MinimaxChatAdapter` (basic config, streaming, temperature clamping, max tokens, missing API key, all kwargs, model variants)
- [x] 3 integration tests for registry pattern matching and end-to-end adapter flow
- [x] 4 provider tests for `_list_minimax_models()` (with/without API key, model names, enum)
- [x] Existing tests pass without regressions (10/10 existing adapter + provider tests)

**To enable:** Add `"MINIMAX_API_KEY": "your-key"` to the Secrets Manager secret (same process as OpenAI/Cohere keys).

### DIFF
--- a/docs/documentation/model-requirements.md
+++ b/docs/documentation/model-requirements.md
@@ -41,7 +41,7 @@ Usually, an `API_KEY` is required to integrate with 3P models. To do so, the [Mo
 
 These keys will be injected at runtime into the Lambda function Environment Variables; they won't be visible in the AWS Lambda Console.
 
-For example, if you wish to be able to interact with AI21 Labs., OpenAI's and Cohere endpoints:
+For example, if you wish to be able to interact with AI21 Labs., OpenAI's, Cohere and MiniMax endpoints:
 
 - Open the [Model Interface Keys Secret](https://github.com/aws-samples/aws-genai-llm-chatbot/blob/main/lib/model-interfaces/langchain/index.ts#L38) in Secrets Manager. You can find the secret name in the stack output, too.
 - Update the Secrets by adding a key to the JSON
@@ -50,7 +50,8 @@ For example, if you wish to be able to interact with AI21 Labs., OpenAI's and Co
 {
   "AI21_API_KEY": "xxxxx",
   "OPENAI_API_KEY": "sk-xxxxxxxxxxxxxxx",
-  "COHERE_API_KEY": "xxxxx"
+  "COHERE_API_KEY": "xxxxx",
+  "MINIMAX_API_KEY": "xxxxx"
 }
 ```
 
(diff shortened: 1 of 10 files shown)


JiwaniZakir (commented):
The regex pattern `r"^minimax*"` in `minimax_chat.py` (registry line) has a subtle but critical bug: the `*` is a quantifier on the preceding character `x` (meaning "zero or more x's"), not a glob wildcard. This means the pattern matches strings like "minima", "minimax", "minimaxxx" — but more importantly, since the registered model IDs are `"MiniMax-M2.7"`, `"MiniMax-M2.5"`, etc. (capital M), a case-sensitive match against `^minimax` will fail entirely and none of these models will ever be routed to `MinimaxChatAdapter`. The pattern should be `r"^MiniMax"` or `r"(?i)^minimax"` with `re.IGNORECASE`.

Additionally, `get_minimax_client()` in `clients.py` is added but never referenced anywhere in the visible diff — the adapter in `minimax_chat.py` reads `os.environ.get("MINIMAX_API_KEY")` directly rather than going through `genai_core.parameters.get_external_api_key()`. This inconsistency means the two code paths behave differently depending on how the key is stored, and the new client function appears to be dead code.

Finally, the `_MINIMAX_MODELS` list in `provider.py` includes a `"description"` field for each model, but the dict comprehension in `_list_minimax_models()` never includes `"description"` in the returned model dict — the field is defined and documented but silently dropped.

octo-patch:
Good catch @JiwaniZakir! Fixed the regex pattern from `r"^minimax*"` to `r"^minimax.*"` — the `*` was incorrectly quantifying the `x` character instead of acting as a wildcard. Thanks for the review!

